PR #81, opened 2018-06-17
Add image tag support for DockerID

nmiyake:
* Change Docker tags in configuration from a list of tags to a
  map of tags
* Add logic that translates the old format (list of tags) to new
  format (map) by using index in list as the key
* Update DockerID to add support for specifying tag name

Taken together, this allows tasks that take DockerIDs (such as
docker build and docker push) to operate on specific tags only.


nmiyake:
@jamesross3 for SA

PR #122, opened 2018-10-08
Add build-args-script to defaultdockerbuilder

### LINKED ISSUE #119: Add `build-args-script` option to add flexibility to docker build arguments
One simple use case: I'd like to specify my base image version in only one place, but I have multiple `Dockerfile`s referenced in my `distgo` configuration. Given a file `base-image.version` with the contents `1.0.0`, I'd like to be able to have my docker command look something like:
`docker build --build-args BASE_IMAGE_VERSION=1.0.0`. A valid approach could look like:

```
products:
  foo:
    # build and dist blocks
    docker:
      docker-builders:
            build-1:
              # other fields...
              config:
                  build-args-script: |
                    echo "--build-args BASE_IMAGE_VERSION=$(cat base-image.version)"
```
@nmiyake

nmiyake:
Provides the ability to specify arguments based on the execution
of a script.

Fixes #119

### DIFF
--- a/dockerbuilder/defaultdockerbuilder/config/internal/v0/config.go
+++ b/dockerbuilder/defaultdockerbuilder/config/internal/v0/config.go
@@ -24,9 +24,9 @@ type Config struct {
 
 	// BuildArgsScript is the content of a script that is written to a file and run before this image is built to
 	// provide supplemental "docker build" arguments for the image. The content of this value is written to a file and
-	// executed. The script process inherits the environment variables of the Go process. Each line of output of the
-	// script is provided to the "docker build" command as a separate argument. The arguments produced by the script are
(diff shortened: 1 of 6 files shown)


dockerbuilder/defaultdockerbuilder/config/internal/v0/config.go
```diff
@@ -21,6 +21,13 @@ import (
 
 type Config struct {
 	BuildArgs []string `yaml:"build-args,omitempty"`
+
+	// BuildArgsScript is the content of a script that is written to a file and run before this image is built to
```
jamesross3: I think it's worth clarifying in this comment what the working directory of the script is; I assume it's wherever `godelw` is, but some people may assume it's `context-dir`.

PR #133, opened 2019-03-28
Add support for attaching properties to Artifactory uploads

jasonma:
This introduces a config in the Artifactory publisher to add properties (https://www.jfrog.com/confluence/display/RTF/Using+Properties+in+Deployment+and+Resolution) to an Artifactory upload.  The property value can be a literal string, or take on the value of an environment variable.


publisher/artifactory/config/properties.go
```diff
@@ -0,0 +1,51 @@
+// Copyright 2016 Palantir Technologies, Inc.
```
nmiyake: Since this is a new file, copyright should be 2019.

I've updated the configuration on the project to fix this (it will fill in current year for new files). Can you update/rebase this PR on latest master and then update this year to 2019?

publisher/artifactory/config/internal/v0/config.go
```diff
@@ -25,6 +25,9 @@ type Config struct {
 	publisher.BasicConnectionInfo `yaml:",inline,omitempty"`
 	Repository                    string `yaml:"repository,omitempty"`
 	NoPOM                         bool   `yaml:"no-pom,omitempty"`
+	// Properties is a map of properties to attach to an artifact on publishing. Environment variables
```
nmiyake: The comment should mention that the values of the map are rendered as Go templates (that way, users know to do special escaping etc. as necessary).

If you don't want to actually support full Go templates for the values (and just want the contract described here), then the processing code should be updated to perform a find/replace on the directive and do the env variable rendering in that manner. Don't have a particularly strong feeling on this either way.

publisher/artifactory/config/properties.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+func EncodeProperties(properties map[string]string) (string, error) {
+	if len(properties) == 0 {
+		return "", nil
+	}
+
+	var encoded []string
+	for k, v := range properties {
+		tmpl, err := template.New("properties").Funcs(template.FuncMap{
```
nmiyake: Do this once before the loop rather than every time through

publisher/artifactory/config/properties.go
```diff
@@ -0,0 +1,51 @@
...
+		}
+		output := &bytes.Buffer{}
+		if err := tmpl.Execute(output, nil); err != nil {
+			return "", errors.Wrapf(err, "failed to execute template")
+		}
+		if len(output.String()) > 0 {
+			encoded = append(encoded, fmt.Sprintf("%s=%s", k, output.String()))
+		}
+	}
+	return strings.Join(encoded, ";"), nil
```
nmiyake: Currently, the output of this function is not deterministic because map iteration order is random. Although this isn't necessarily bad, in general I think it's better for things to be deterministic because it can help with debugging/caching/etc.

Would recommend sorting `encoded` first to make output deterministic.

Also, per other comment below, recommend just `return encoded, nil` as the API (after sorting)

publisher/artifactory/config/properties.go
```diff
@@ -0,0 +1,51 @@
...
+		}).Parse(v)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to parse template")
+		}
+		output := &bytes.Buffer{}
+		if err := tmpl.Execute(output, nil); err != nil {
+			return "", errors.Wrapf(err, "failed to execute template")
+		}
+		if len(output.String()) > 0 {
+			encoded = append(encoded, fmt.Sprintf("%s=%s", k, output.String()))
```
nmiyake: nit, but this calls `output.String()` twice. Would recommend:

```go
if outStr := output.String(); len(outStr) > 0 {
   .... k, outStr));
}
```

publisher/artifactory/config/properties.go
```diff
@@ -0,0 +1,51 @@
...
+	"bytes"
+	"fmt"
+	"html/template"
+	"os"
+	"strings"
+
+	"github.com/pkg/errors"
+)
+
+func EncodeProperties(properties map[string]string) (string, error) {
```
nmiyake: Please add comment to the function that describes what it does. Would also recommend having this return `[]string`, where contract is that this function takes the map, renders values as Go templates (with `env` function), then returns sorted slice of strings of the form `key=renderedVal`

publisher/artifactory/config/properties.go
nmiyake: After reviewing further, not sure that it makes sense for this function to be in a separate file in this package -- it's not directly related to configuration.

I would just declare this new function near where it's used.

publisher/artifactory/publisher.go
```diff
@@ -216,3 +220,15 @@ func (p *artifactoryPublisher) artifactorySetSHA256Checksum(cfg config.Artifacto
 	}
 	return nil
 }
+
+func (p *artifactoryPublisher) getDeploymentURL(cfg config.Artifactory) (string, error) {
+	url := strings.Join([]string{cfg.URL, "artifactory", cfg.Repository}, "/")
+	if len(cfg.Properties) > 0 {
+		encoded, err := config.EncodeProperties(cfg.Properties)
```
nmiyake: Per feedback above, would change this function to return `[]string` and then perform the join with `;` at this level. Main reason is that right now the `;` appending that happens in this block depends on the `;` separation done by the function -- I think it's cleaner for the function to focus on the render/transform, and then this function is concerned with stringing all the parts together in a way that makes sense to Artifactory.

publisher/artifactory/config/properties.go
jasonma: Done.

publisher/artifactory/config/properties.go
jasonma: Done.

publisher/artifactory/config/properties.go
jasonma: I can't call `Parse` on a template after calling `Execute` on it, so I'm cloning the template.  Let me know if there's a better approach here.

publisher/artifactory/config/properties.go
jasonma: Done.

publisher/artifactory/config/internal/v0/config.go
jasonma: Done.

publisher/artifactory/config/properties.go
jasonma: Done.

publisher/artifactory/publisher.go
jasonma: Done.

publisher/artifactory/config/internal/v0/config.go
```diff
@@ -25,6 +25,10 @@ type Config struct {
 	publisher.BasicConnectionInfo `yaml:",inline,omitempty"`
 	Repository                    string `yaml:"repository,omitempty"`
 	NoPOM                         bool   `yaml:"no-pom,omitempty"`
+	// Properties is a map of properties to attach to an artifact on publishing. The value can be a
+	// string literal or a Go template. In particular, it is possible to get the value of a environment
```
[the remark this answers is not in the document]
nmiyake: Also, can you include a link to the description of what Artifactory properties are so that it's clear that's what's being referenced here?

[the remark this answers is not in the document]
jasonma: Done.

publisher/artifactory/publisher.go
```diff
@@ -216,3 +224,48 @@ func (p *artifactoryPublisher) artifactorySetSHA256Checksum(cfg config.Artifacto
...
+// encodeProperties takes in a map[string]string of properties, renders each value as a Go template, and returns
+// the sorted slice of strings of the form `key=renderedVal`. The Go template can include the `env` function,
+// which fetches an environment variable.
+func encodeProperties(properties map[string]string) ([]string, error) {
+	if len(properties) == 0 {
+		return nil, nil
+	}
+
+	tmpl := template.New("properties").Funcs(template.FuncMap{
+		"env": func(key string) string {
```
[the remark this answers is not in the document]
jasonma: Done.

publisher/artifactory/integration_test/integration_test.go
```diff
@@ -244,6 +244,47 @@ products:
...
+      info:
+        artifactory:
+          config:
+            url: http://artifactory.domain.com
+            username: testUsername
+            password: testPassword
+            repository: testRepo
+            properties:
+              key1: value1
+              key2: value2
```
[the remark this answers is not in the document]
jasonma: Yep, good flag -- done.

publisher/artifactory/integration_test/integration_test.go
```diff
@@ -244,8 +245,104 @@ products:
...
+				WantOutput: func(projectDir string) string {
+					return fmt.Sprintf(`[DRY RUN] Uploading out/dist/foo/1.0.0/os-arch-bin/foo-1.0.0-%s.tgz to http://artifactory.domain.com/artifactory/testRepo;key1=value1;key2=value2/com/test/group/foo/1.0.0/foo-1.0.0-%s.tgz
+[DRY RUN] Uploading to http://artifactory.domain.com/artifactory/testRepo;key1=value1;key2=value2/com/test/group/foo/1.0.0/foo-1.0.0.pom
+`, osarch.Current().String(), osarch.Current().String())
+				},
+			},
+		},
+	)
+
+	err = os.Setenv("TEST_ENV_VAR", "testValue")
```
nmiyake: I don't think it makes sense to mingle this standalone test with the other test-case based tests.

Can you just define a dedicated test function for this (something along the lines of `TestArtifactoryPublisherRendersGoTemplates`)?

The other option would be to set the environment variable at the beginning of this test and then just add the test case, but that introduces an unnatural coupling between setting the environment variable and the specific test case, so having the stand-alone test probably makes the most sense.

nmiyake:
Final piece of feedback was very minor, so I made the change and pushed as #135. Thanks for implementing this!

jasonma:
@nmiyake Sounds great -- thanks for reviewing this!

PR #143, opened 2019-08-01
Write dist scripts to the dist working directory.

### LINKED ISSUE #142: dist script will dirty the build
When using godel's dist command/plugin (which calls into distgo) we hit an issue if using a dist script. The build is marked `dirty` even if it isn't which can lead to artifacts being missed and therefore a failed build.

Example:

        test-app:
          type: manual
          config:
            extension: tgz
          script: |
            #!/bin/bash
            set -euo pipefail
            # jump out of yaml to the dist-script
            ${PROJECT_DIR}/dist-scripts/dist.sh

Our `dist.sh` uses godel's `godelw project-version` which is the version used for the artifact existence check.

The problem is that distgo writes a script file _inside_ the `PROJECT_DIR` which dirties the git workspace. See https://github.com/palantir/distgo/blob/master/distgo/script.go#L30. This script should be written to a build directory to avoid dirtying the workspace and messing with the version.

The issue is particularly nasty to chase down as the workspace is clean when inspecting what is dirty. The reason for _that_ is the script file is cleaned up after use and disappears by the time an engineer can look at what went wrong.

aggie05engineer:
Previously, all scripts were writen to the PROJECT_DIR. This ends
up changing the git workspace and "dirties" the environment. This in
turn ends up breaking tagged builds that rely on non-dirty versions.

The fix isn't pretty and likely needs a better refactor than re-using
ProjectInfo.ProjectDir everywhere.

Fixes #142.

### DIFF
--- a/distgo/build/build.go
+++ b/distgo/build/build.go
@@ -72,7 +72,7 @@ func Run(projectInfo distgo.ProjectInfo, productParams []distgo.ProductParam, bu
 		}
 
 		// execute build script
-		if err := distgo.WriteAndExecuteScript(projectInfo, currProductParam.Build.Script, distgo.BuildScriptEnvVariables(currProductTaskOutputInfo), stdout); err != nil {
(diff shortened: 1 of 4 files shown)


aggie05engineer:
The dist script sees a dirty git workspace but the use of VERSION in the build script fixes what I needed. This particular branch is likely not worth merging at this point.

PR #189, opened 2020-11-16
Overwrite existing dists

### LINKED ISSUE #190: Running build+dist multiple times causes failures for the os-arch-bin and bin disters
## What happened?
Running build+dist multiple times in sequence for the `bin` and `os-arch-bin` disters causes failures due to the tar.gz archive already existing.
The archiving package we use to create the tar.gz ([github.com/mholt/archiver](github.com/mholt/archiver)) used to overwrite the archive if it existed. This behavior changed when we updated to the v3 version of the package: https://github.com/palantir/distgo/pull/153. The new behavior now returns an error if the file exists, which we did not catch as part of the module update. It is still possible to configure the archiver to overwrite files if they exist, which if set, would bring us back to the old/intended behavior: https://github.com/mholt/archiver/blob/25e050dbb4e3c591402916201098241bb661a2f7/tar.go#L21

This can be easily reproduced by running build + dist twice and observing an error on the second dist call.

## What did you want to happen?

Ideally the archive will be overwritten if it already exists, rather than error out.

tabboud:
## Before this PR
The [v3 update](https://github.com/palantir/distgo/pull/153) for the archiver pkg now returns errors if the archive already exists
https://github.com/mholt/archiver/blob/25e050dbb4e3c591402916201098241bb661a2f7/tar.go#L89. Before the v3 bump, files were overwritten by default. The new behavior exposes a race condition between build and dist, which can be seen by running the added test case.

## After this PR
==COMMIT_MSG==
Make os-arch-bin and bin disters overwrite destination archive if it already exists.
==COMMIT_MSG==

Fixes #190

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/189)


nmiyake (changes requested):
Thanks for diagnosing this issue and submitting this fix! Provided some feedback.

> Reviewed 4 of 4 files at r1.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189#-:-MMI96gp72oyb2Qx05Hq:b-f945lo)* status: all files reviewed, 3 unresolved discussions (waiting on @tabboud)

___
*[dister/bin/dister.go, line 80 at r1](https://reviewable.io/reviews/palantir/distgo/189#-MMI7bIDD_HoJPC9DeBd:-MMI7bIDD_HoJPC9DeBe:br7c91e) ([raw file](https://github.com/palantir/distgo/blob/fb8a11b4de3c122cc585b81da1eae790a5fd484f/dister/bin/dister.go#L80)):*
> ```Go
> 	distWorkDir := productTaskOutputInfo.ProductDistWorkDirs()[distID]
> 	dstPath := productTaskOutputInfo.ProductDistArtifactPaths()[distID][0]
> 	tarGZ := archiver.DefaultTarGz
> ```

This should be `tarGZ := archiver.NewTarGz()` -- otherwise, this modifies the global `archier.DefaultTarGz`, which may have unintended consequences.
___
*[dister/bin/integration_test/integration_test.go, line 32 at r1](https://reviewable.io/reviews/palantir/distgo/189#-MMI8zUq3CA5k0fyPCF9:-MMI8zUq3CA5k0fyPCFA:blyt0jt) ([raw file](https://github.com/palantir/distgo/blob/fb8a11b4de3c122cc585b81da1eae790a5fd484f/dister/bin/integration_test/integration_test.go#L32)):*
> ```Go
> )
> 
> func TestBuildDistTwice(t *testing.T) {
> ```

If doing this, should have the same test for `os-arch-bin` as well.
___
*[dister/osarchbin/dister.go, line 115 at r1](https://reviewable.io/reviews/palantir/distgo/189#-MMI7u9y0MjDaJoXuRHO:-MMI7u9y0MjDaJoXuRHP:b-4p6l9x) ([raw file](https://github.com/palantir/distgo/blob/fb8a11b4de3c122cc585b81da1eae790a5fd484f/dister/osarchbin/dister.go#L115)):*
> ```Go
> 			itemPaths[i] = filepath.Join(workDir, item.Name())
> 		}
> 		tarGZ := archiver.DefaultTarGz
> ```

Same feedback as for bin dister

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: 0 of 4 files reviewed, 3 unresolved discussions (waiting on @nmiyake)

___
*[dister/bin/dister.go, line 80 at r1](https://reviewable.io/reviews/palantir/distgo/189#-MMI7bIDD_HoJPC9DeBd:-MMIKpcZA6wavGj11kHt:b-yy10zm) ([raw file](https://github.com/palantir/distgo/blob/fb8a11b4de3c122cc585b81da1eae790a5fd484f/dister/bin/dister.go#L80)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

This should be `tarGZ := archiver.NewTarGz()` -- otherwise, this modifies the global `archier.DefaultTarGz`, which may have unintended consequences.
</blockquote></details>

good catch! done
___
*[dister/bin/integration_test/integration_test.go, line 32 at r1](https://reviewable.io/reviews/palantir/distgo/189#-MMI8zUq3CA5k0fyPCF9:-MMIKrU87Y-zCNJFrR8c:bi3omqs) ([raw file](https://github.com/palantir/distgo/blob/fb8a11b4de3c122cc585b81da1eae790a5fd484f/dister/bin/integration_test/integration_test.go#L32)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

If doing this, should have the same test for `os-arch-bin` as well.
</blockquote></details>

I decided to move this test out of the full integration test suite and into a targeted unit test that handles both `os-arch-bin` and `bin` disters. Let me know what you think.
___
*[dister/osarchbin/dister.go, line 115 at r1](https://reviewable.io/reviews/palantir/distgo/189#-MMI7u9y0MjDaJoXuRHO:-MMIL4jc6Hl9vAdYNSzE:b-896fix) ([raw file](https://github.com/palantir/distgo/blob/fb8a11b4de3c122cc585b81da1eae790a5fd484f/dister/osarchbin/dister.go#L115)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Same feedback as for bin dister
</blockquote></details>

Done.

nmiyake (changes requested):
Thanks! Feedback around changelog/commit message and minor thing for variable/const declaration in test.

> Reviewed 5 of 6 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189#-:-MMMfuJ176T2LP_69QkY:b-atu3m0)* status: all files reviewed, 2 unresolved discussions (waiting on @tabboud)

___
*[changelog/@unreleased/pr-189.v2.yml, line 4 at r2](https://reviewable.io/reviews/palantir/distgo/189#-MMMRyyt90FRmOg668cf:-MMMRyyt90FRmOg668cg:bkmyi9o) ([raw file](https://github.com/palantir/distgo/blob/73eae2421aff8a9afbaf3a5f505a17935266cf35/changelog/@unreleased/pr-189.v2.yml#L4)):*
> ```YAML
> fix:
>   description: |
>     overwrite existing dists
> ```

Can you update this description to make it more complete?

Would be helpful to:
1. Open new issue that describes the bug (running "build + dist" twice for bin dister and os-arch-bin dister causes failures)
2. Expand this description: "Make it so that bin and os-arch-bin disters will overwrite destination file if it already exists. Fixes {issueRef}."
___
*[distgo/dist/dist_test.go, line 521 at r2](https://reviewable.io/reviews/palantir/distgo/189#-MMMSmRuCfPTfaBHGOHt:-MMMSmRv7xO-NgNbW8j5:b-ft9u5) ([raw file](https://github.com/palantir/distgo/blob/73eae2421aff8a9afbaf3a5f505a17935266cf35/distgo/dist/dist_test.go#L521)):*
> ```Go
> 
> 			var (
> 				distRuns = 2
> ```

nit: can we make this:

```
const distRuns = 2
var modTime time.Time
```

Although it doesn't matter in practice, declaring `distRuns` as a const clearly signals intent that this is a compile-time constant.

nmiyake (changes requested):
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: all files reviewed, 3 unresolved discussions (waiting on @tabboud)

___
*[distgo/dist/dist_test.go, line 528 at r2](https://reviewable.io/reviews/palantir/distgo/189#-MMN6uRI3LafgmiRNti6:-MMN6uRJ0QCSQ_-UkTQI:bhtpv23) ([raw file](https://github.com/palantir/distgo/blob/73eae2421aff8a9afbaf3a5f505a17935266cf35/distgo/dist/dist_test.go#L528)):*
> ```Go
> 				require.NoError(t, err)
> 
> 				info, err := os.Stat(path.Join(projectDir, "out", "dist", "foo", "unspecified", "os-arch-bin", fmt.Sprintf("foo-unspecified-%s.tgz", osarch.Current().String())))
> ```

Why is "os-arch-bin" hard-coded here? Can you verify that this works if you comment out that dister and just test the "bin" dister?

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: 2 of 4 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[changelog/@unreleased/pr-189.v2.yml, line 4 at r2](https://reviewable.io/reviews/palantir/distgo/189#-MMMRyyt90FRmOg668cf:-MMNESdR7Aku8KrzYcjE:b-896fix) ([raw file](https://github.com/palantir/distgo/blob/73eae2421aff8a9afbaf3a5f505a17935266cf35/changelog/@unreleased/pr-189.v2.yml#L4)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Can you update this description to make it more complete?

Would be helpful to:
1. Open new issue that describes the bug (running "build + dist" twice for bin dister and os-arch-bin dister causes failures)
2. Expand this description: "Make it so that bin and os-arch-bin disters will overwrite destination file if it already exists. Fixes {issueRef}."
</blockquote></details>

Done.
___
*[distgo/dist/dist_test.go, line 528 at r2](https://reviewable.io/reviews/palantir/distgo/189#-MMN6uRI3LafgmiRNti6:-MMNEUci-mYiQTr-N5b1:banfnc1) ([raw file](https://github.com/palantir/distgo/blob/73eae2421aff8a9afbaf3a5f505a17935266cf35/distgo/dist/dist_test.go#L528)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Why is "os-arch-bin" hard-coded here? Can you verify that this works if you comment out that dister and just test the "bin" dister?
</blockquote></details>

great catch! I also wasn't aware the the `testfuncs.NewProjectParam()` added in the os-arch-bin dister by default, so I opted for a test case function that returns the appropriate file mod time for each of the disters. I also scoped down the test case config.

nmiyake (changes requested):
> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: all files reviewed, 2 unresolved discussions (waiting on @tabboud)

___
*[changelog/@unreleased/pr-189.v2.yml, line 5 at r3](https://reviewable.io/reviews/palantir/distgo/189#-MMNfk6Y0RkrC9R7isI_:-MMNfk6Y0RkrC9R7isIa:b2q5d6x) ([raw file](https://github.com/palantir/distgo/blob/550e6d9d5e8a343e21ca9cda36bddd892d76de39/changelog/@unreleased/pr-189.v2.yml#L5)):*
> ```YAML
>   description: |
>     Make it so the os-arch-bin and bin disters will overwrite the destination file if it exists.
>     Fixes https://github.com/palantir/distgo/issues/190
> ```

This should be "Fixes #190" -- this will make GitHub auto-close the issue when this PR merges (and it will take care of having the proper link in the UI).
___
*[distgo/dist/dist_test.go, line 472 at r3](https://reviewable.io/reviews/palantir/distgo/189#-MMNfqwX8TIHShjkEe4Q:-MMNfqwX8TIHShjkEe4R:b-atgz32) ([raw file](https://github.com/palantir/distgo/blob/550e6d9d5e8a343e21ca9cda36bddd892d76de39/distgo/dist/dist_test.go#L472)):*
> ```Go
> 
> // TestDistOverwrites verifies that subsequent runs of the dist task will overwrite the artifacts.
> func TestDistOverwrites(t *testing.T) {
> ```

After taking a look at another custom dister with this issue, I think it's worth stepping back and thinking about the bigger picture.

In identifying this issue and the fix, we've basically identified that part of the contract of dist generation should be that previous artifacts should be overwritten (or, more precisely, multiple generations should not fail).

The approach of the fix in the current PR is that it's the responsibility of individual disters to make sure that regeneration succeeds. If that's the case, then each dister needs a way to test/verify this, so I think we should take the approach in the first draft of your PR and add a new function to `distertester` that runs the logic of doing `dist` twice after a `build` so that other custom disters can verify their regeneration functionality.

The other approach would be to say that, since this expectation is common across disters, `distgo` should handle it -- that is, the logic should do something like remove the declared dist artifacts before it executes the "dist" task. However, this seems a bit more dangerous and is a bigger change, so not sure that it's worth making.

Interested to hear your thoughts -- I think I have a slight lean towards the first approach right now just because I think it's safer, but I do worry a bit that it's up to individual dister implementers to know the contract of being able to handle the output already existing and testing using the provided library.

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: 3 of 4 files reviewed, 2 unresolved discussions (waiting on @nmiyake)

___
*[changelog/@unreleased/pr-189.v2.yml, line 5 at r3](https://reviewable.io/reviews/palantir/distgo/189#-MMNfk6Y0RkrC9R7isI_:-MMRBc8k8gpNESpIk2O-:bemyt4x) ([raw file](https://github.com/palantir/distgo/blob/550e6d9d5e8a343e21ca9cda36bddd892d76de39/changelog/@unreleased/pr-189.v2.yml#L5)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

This should be "Fixes #190" -- this will make GitHub auto-close the issue when this PR merges (and it will take care of having the proper link in the UI).
</blockquote></details>

oh cool, thats good to know!
___
*[distgo/dist/dist_test.go, line 472 at r3](https://reviewable.io/reviews/palantir/distgo/189#-MMNfqwX8TIHShjkEe4Q:-MMRDu4b3CFWuv_I5XNi:b-ko5vjp) ([raw file](https://github.com/palantir/distgo/blob/550e6d9d5e8a343e21ca9cda36bddd892d76de39/distgo/dist/dist_test.go#L472)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

After taking a look at another custom dister with this issue, I think it's worth stepping back and thinking about the bigger picture.

In identifying this issue and the fix, we've basically identified that part of the contract of dist generation should be that previous artifacts should be overwritten (or, more precisely, multiple generations should not fail).

The approach of the fix in the current PR is that it's the responsibility of individual disters to make sure that regeneration succeeds. If that's the case, then each dister needs a way to test/verify this, so I think we should take the approach in the first draft of your PR and add a new function to `distertester` that runs the logic of doing `dist` twice after a `build` so that other custom disters can verify their regeneration functionality.

The other approach would be to say that, since this expectation is common across disters, `distgo` should handle it -- that is, the logic should do something like remove the declared dist artifacts before it executes the "dist" task. However, this seems a bit more dangerous and is a bigger change, so not sure that it's worth making.

Interested to hear your thoughts -- I think I have a slight lean towards the first approach right now just because I think it's safer, but I do worry a bit that it's up to individual dister implementers to know the contract of being able to handle the output already existing and testing using the provided library.
</blockquote></details>

This is a great point, especially since the logic to test another custom dister will likely be exactly the same as this test, so it seems reasonable to create public testing APIs for all disters to use for verifying their behavior.

I am also leaning towards the first approach, as it seems reasonable that the overwriting (or cleanup) of the generated dist is handled by the custom dister itself, rather than move that logic into distgo. The only concern I have is that there is no way to enforce this implicit contract on all disters outside of them using the public testing methods and us providing documentation. One alternative, which I don't think we should go with, would be to update the `Dister` interface to add a `Cleanup()` or similar method that all disters must implement. Then from within `RunDist` we can call out to `Cleanup()`. The extra interface method coupled with another public testing api would enforce all disters abide by the contract.

I will go ahead and add in the public `distertester` test case in a pre-commit, since we will want this regardless, and then wire the usage through into this PR.

nmiyake (changes requested):
> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: all files reviewed, 1 unresolved discussion (waiting on @tabboud)

___
*[distgo/dist/dist_test.go, line 472 at r3](https://reviewable.io/reviews/palantir/distgo/189#-MMNfqwX8TIHShjkEe4Q:-MMRre437yUehujbJyCj:b-alriam) ([raw file](https://github.com/palantir/distgo/blob/550e6d9d5e8a343e21ca9cda36bddd892d76de39/distgo/dist/dist_test.go#L472)):*
<details><summary><i>Previously, tabboud (Tony Abboud) wrote…</i></summary><blockquote>

This is a great point, especially since the logic to test another custom dister will likely be exactly the same as this test, so it seems reasonable to create public testing APIs for all disters to use for verifying their behavior.

I am also leaning towards the first approach, as it seems reasonable that the overwriting (or cleanup) of the generated dist is handled by the custom dister itself, rather than move that logic into distgo. The only concern I have is that there is no way to enforce this implicit contract on all disters outside of them using the public testing methods and us providing documentation. One alternative, which I don't think we should go with, would be to update the `Dister` interface to add a `Cleanup()` or similar method that all disters must implement. Then from within `RunDist` we can call out to `Cleanup()`. The extra interface method coupled with another public testing api would enforce all disters abide by the contract.

I will go ahead and add in the public `distertester` test case in a pre-commit, since we will want this regardless, and then wire the usage through into this PR.
</blockquote></details>

Thanks -- for now let's just add the common test without modifying the API, since that should probably be generally useful. If we were going to make the API change you described I think that would require formalizing some other API points as well (for example, are all dist artifacts expected to be files, or is it OK for it to produce directories? If the latter, how are those expected to be handled? Etc.).

distgo/dist/dist_test.go
```diff
@@ -466,6 +472,57 @@ func main() {}
 	}
 }
 
+// TestRepeatedDist verifies that subsequent runs of the dist task will succeed without error for all built-in disters.
+func TestRepeatedDist(t *testing.T) {
```
tabboud, line 476: This is a mix between an integration and unit test, so I am happy to re-locate this if you think it should live elsewhere.
Also I added in the manual dister as well so we have full coverage on the built-in disters

nmiyake (approved):
> Reviewed 1 of 1 files at r5.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/189)* status: all files reviewed, 2 unresolved discussions (waiting on @tabboud)

___
*[distgo/dist/dist_test.go, line 476 at r5](https://reviewable.io/reviews/palantir/distgo/189#-MMIIQJKSTF92rM3px2c-r5-476:-MMwGw7bAPDjkp-JHkI7:b-wpj8to) ([raw file](https://github.com/palantir/distgo/blob/e2172521556207510586b400457d5c38a2cf5c05/distgo/dist/dist_test.go#L476)):*
<details><summary><i>Previously, tabboud (Tony Abboud) wrote…</i></summary><blockquote>

This is a mix between an integration and unit test, so I am happy to re-locate this if you think it should live elsewhere.
Also I added in the manual dister as well so we have full coverage on the built-in disters
</blockquote></details>

Looks great -- thank you!

PR #191, opened 2020-11-19
add public test to verify dist overwrites

tabboud:
## Before this PR
There are no public test functions to verify that all disters will be able to run build+dist multiple times with success. We have found that the `bin` and `os-arch-bin` disters do not currently succeed when run multiple times, which will be fixed in #189. However, any other dister outside of the built in ones should also be able to verify this behavior. Also related to the following issue: https://github.com/palantir/distgo/issues/190

## After this PR
==COMMIT_MSG==
Add public test to verify dist overwrites
==COMMIT_MSG==

The added public test verifies that running dist twice will succeed without error for the provided dister config.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/191)

### DIFF
--- a/dister/distertester/distertester.go
+++ b/dister/distertester/distertester.go
@@ -156,8 +156,8 @@ func RunAssetDistTest(t *testing.T,
 	}
 }
 
-// RunDistOverwritesTest verifies that running build+dist multiple times, for the provided distType,
-// will succeed without error.
+// RunDistOverwritesTest verifies that running the "dist" task multiple times with
+// the provided DistersConfig will succeed without error.
 func RunDistOverwritesTest(t *testing.T,
 	pluginProvider pluginapitester.PluginProvider,
 	assetProvider pluginapitester.AssetProvider,
--- a/dister/distertester/distertester.go
+++ b/dister/distertester/distertester.go
@@ -16,13 +16,13 @@ package distertester
 
 import (
 	"bytes"
+	"fmt"
 	"io/ioutil"
 	"os"
 	"path"
 	"path/filepath"
 	"sort"
 	"testing"
-	"time"
 
 	"github.com/nmiyake/pkg/dirs"
 	"github.com/nmiyake/pkg/gofiles"
@@ -161,22 +161,21 @@ func RunAssetDistTest(t *testing.T,
 func RunDistOverwritesTest(t *testing.T,
 	pluginProvider pluginapitester.PluginProvider,
 	assetProvider pluginapitester.AssetProvider,
-	distType string,
+	distersCfg distgoconfig.DistersConfig,
 ) {
-	osarches := []osarch.OSArch{osarch.Current()}
+	var (
+		productName = "dist-overwrite-test-product"
+		osarches    = []osarch.OSArch{osarch.Current()}
+	)
+
 	projectCfg := distgoconfig.ProjectConfig{
 		Products: distgoconfig.ToProductsMap(map[distgo.ProductID]distgoconfig.ProductConfig{
-			"foo": {
+			distgo.ProductID(productName): {
 				Build: distgoconfig.ToBuildConfig(&distgoconfig.BuildConfig{
-					MainPkg: stringPtr("foo"),
 					OSArchs: &osarches,
 				}),
 				Dist: distgoconfig.ToDistConfig(&distgoconfig.DistConfig{
-					Disters: distgoconfig.ToDistersConfig(&distgoconfig.DistersConfig{
-						distgo.DistID(distType): {
-							Type: stringPtr(distType),
-						},
-					}),
+					Disters: distgoconfig.ToDistersConfig(&distersCfg),
 				}),
 			},
 		}),
@@ -185,10 +184,8 @@ func RunDistOverwritesTest(t *testing.T,
 	require.NoError(t, err)
 	internalFiles := map[string][]byte{
 		"godel/config/dist-plugin.yml": projectCfgBytes,
-		"foo/foo.go":                   []byte(`package main; func main() {}`),
-		"go.mod": []byte(`module foo
-go 1.15
-`),
+		"main.go":                      []byte(`package main; func main() {}`),
+		"go.mod":                       []byte(fmt.Sprintf("module %s\ngo 1.15", productName)),
 	}
 	wd, err := os.Getwd()
 	require.NoError(t, err)
@@ -229,32 +226,23 @@ go 1.15
 		assetProviders = append(assetProviders, assetProvider)
 	}
 
-	for j := 0; j < 2; j++ {
-		// run build task
-		buildBuf := new(bytes.Buffer)
-		_, err := pluginapitester.RunPlugin(
-			pluginProvider,
-			assetProviders,
-			"build", nil,
-			projectDir, false, buildBuf)
-		require.NoError(t, err, buildBuf.String())
+	// run build task once
+	buildBuf := new(bytes.Buffer)
+	_, err = pluginapitester.RunPlugin(
+		pluginProvider,
+		assetProviders,
+		"build", nil,
+		projectDir, false, buildBuf)
+	require.NoError(t, err, buildBuf.String())
 
-		// run dist task
+	for j := 0; j < 2; j++ {
+		// run dist task twice and ensure no errors
 		distBuf := new(bytes.Buffer)
 		_, err = pluginapitester.RunPlugin(
 			pluginProvider,
 			assetProviders,
-			"dist", nil,
+			"dist", []string{"--force"}, // use --force to skip build caching
 			projectDir, false, distBuf)
 		require.NoError(t, err, distBuf.String())
-
-		if j == 0 {
-			// wait some time before regeneration
-			time.Sleep(time.Second)
-		}
 	}
-}
-
-func stringPtr(s string) *string {
-	return &s
 }
--- a/dister/distertester/distertester.go
+++ b/dister/distertester/distertester.go
@@ -156,21 +156,21 @@ func RunAssetDistTest(t *testing.T,
 	}
 }
 
-// RunDistOverwritesTest verifies that running the "dist" task multiple times with
+// RunRepeatedDistTest verifies that running the "dist" task multiple times with
 // the provided DistersConfig will succeed without error.
-func RunDistOverwritesTest(t *testing.T,
+// This test generates a single build artifact and runs the "dist" task in a way that ignores the build cache
+// to verify the behavior of strictly running "dist" multiple times.
+func RunRepeatedDistTest(t *testing.T,
 	pluginProvider pluginapitester.PluginProvider,
 	assetProvider pluginapitester.AssetProvider,
 	distersCfg distgoconfig.DistersConfig,
 ) {
-	var (
-		productName = "dist-overwrite-test-product"
-		osarches    = []osarch.OSArch{osarch.Current()}
-	)
+	const productName = "dist-overwrite-test-product"
+	osarches    := []osarch.OSArch{osarch.Current()}
 
 	projectCfg := distgoconfig.ProjectConfig{
 		Products: distgoconfig.ToProductsMap(map[distgo.ProductID]distgoconfig.ProductConfig{
-			distgo.ProductID(productName): {
+			productName: {
 				Build: distgoconfig.ToBuildConfig(&distgoconfig.BuildConfig{
 					OSArchs: &osarches,
 				}),
--- a/dister/distertester/distertester.go
+++ b/dister/distertester/distertester.go
@@ -22,13 +22,18 @@ import (
 	"path/filepath"
 	"sort"
 	"testing"
+	"time"
 
 	"github.com/nmiyake/pkg/dirs"
 	"github.com/nmiyake/pkg/gofiles"
+	"github.com/palantir/distgo/distgo"
+	distgoconfig "github.com/palantir/distgo/distgo/config"
 	"github.com/palantir/godel/v2/framework/pluginapitester"
+	"github.com/palantir/godel/v2/pkg/osarch"
 	"github.com/palantir/pkg/gittest"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v2"
 )
 
 type TestCase struct {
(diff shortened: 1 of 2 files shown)


tabboud:
Once this merges, we can add a test case similar to the following for the internal disters:
```go
func TestBuildDistTwice(t *testing.T) {
       pluginPath, err := products.Bin("dist-plugin")
       require.NoError(t, err)
       pluginProvider := pluginapitester.NewPluginProvider(pluginPath)
       distertester.RunDistOverwritesTest(t, pluginProvider, nil, bin.TypeName)
}
```

dister/distertester/distertester.go
```diff
@@ -150,3 +155,106 @@ func RunAssetDistTest(t *testing.T,
...
+
+// RunDistOverwritesTest verifies that running build+dist multiple times, for the provided distType,
+// will succeed without error.
+func RunDistOverwritesTest(t *testing.T,
+	pluginProvider pluginapitester.PluginProvider,
+	assetProvider pluginapitester.AssetProvider,
+	distType string,
+) {
+	osarches := []osarch.OSArch{osarch.Current()}
+	projectCfg := distgoconfig.ProjectConfig{
```
tabboud, line 167: I am particularly interested if you have feedback on this portion. Contrasting the test above, I decided to keep all of this hardcoded and internal to the test case, since we only really care about which dister to use for running the test. I am happy to change it out to make it configurable by the user, similar to the `TestCase` param above.

dister/distertester/distertester.go
```diff
@@ -150,3 +155,106 @@ func RunAssetDistTest(t *testing.T,
...
+		_, err = pluginapitester.RunPlugin(
+			pluginProvider,
+			assetProviders,
+			"dist", nil,
+			projectDir, false, distBuf)
+		require.NoError(t, err, distBuf.String())
+
+		if j == 0 {
+			// wait some time before regeneration
+			time.Sleep(time.Second)
```
tabboud, line 253: I was seeing some test flakes without this and with some very small sleep times, which I believe may be caused from the build + dist running too quickly in succession.

nmiyake (changes requested):
Thanks for breaking this out! Provided some feedback.

> Reviewed 1 of 1 files at r1, 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/191#-:-MMb-LJGDJXF6_YsNkBY:b-fr784l)* status: all files reviewed, 7 unresolved discussions (waiting on @tabboud)

___
*[changelog/@unreleased/pr-191.v2.yml, line 4 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMb-Ut-3IMkOfswwFCb:-MMb-Ut-3IMkOfswwFCc:bp95owk) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/changelog/@unreleased/pr-191.v2.yml#L4)):*
> ```YAML
> improvement:
>   description: |
>     Adds a public test to verify that running build+dist twice succeeds without error.
> ```

Can you make this changelog more precise/descriptive? For example:

```
Adds the distertester.RunDistOverwritesTest function. This function can be used by custom dister implementations to verify that the "dist" operation does not fail when it has to overwrite previous output.
```

I think something like the above makes the change much more legible/understandable when reviewing release notes.
___
*[dister/distertester/distertester.go, line 167 at r1](https://reviewable.io/reviews/palantir/distgo/191#-MMXIjtF9N17WuJ3Avqi-r1-167:-MMauCIq-HFFdvGZi_WX:bybfmej) ([raw file](https://github.com/palantir/distgo/blob/3d760f62566fdb12b04823525b7c565bb638575c/dister/distertester/distertester.go#L167)):*
<details><summary><i>Previously, tabboud (Tony Abboud) wrote…</i></summary><blockquote>

I am particularly interested if you have feedback on this portion. Contrasting the test above, I decided to keep all of this hardcoded and internal to the test case, since we only really care about which dister to use for running the test. I am happy to change it out to make it configurable by the user, similar to the `TestCase` param above.
</blockquote></details>

Thanks for the explanation! Agree with the sentiment that, since this test is meant to be more targeted, it doesn't need to have as much customizability as the other one -- if every client is going to have to repeat some amount of code to run the test, then centralizing that is a good thing.

However, my suggestion would be to change the provided input from `distType string` to `distgoconfig.DistersConfig` instead -- main thinking here is that a dister may require a certain amount of config, so allowing the config for it to be specified would be helpful.
___
*[dister/distertester/distertester.go, line 253 at r1](https://reviewable.io/reviews/palantir/distgo/191#-MMXIjtF9N17WuJ3Avqi-r1-253:-MMazxd5AzsvINm7Wgc_:b-wlnh2b) ([raw file](https://github.com/palantir/distgo/blob/3d760f62566fdb12b04823525b7c565bb638575c/dister/distertester/distertester.go#L253)):*
<details><summary><i>Previously, tabboud (Tony Abboud) wrote…</i></summary><blockquote>

I was seeing some test flakes without this and with some very small sleep times, which I believe may be caused from the build + dist running too quickly in succession.
</blockquote></details>

Got it -- maybe it has to do with the timestamps used for caching. Would you mind checking to see if there's a way to trigger a "dist" that always makes it run the build? Or if having a dist that doesn't use build output does so? If we could find that, then we could test that directly rather than using an approach like this.
___
*[dister/distertester/distertester.go, line 169 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMauxR7201BSX6gf9QS:-MMauxR7201BSX6gf9QT:b-i11l89) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L169)):*
> ```Go
> 	projectCfg := distgoconfig.ProjectConfig{
> 		Products: distgoconfig.ToProductsMap(map[distgo.ProductID]distgoconfig.ProductConfig{
> 			"foo": {
> ```

I know we often use this naming for tests we write, but since this is part of a test API that others will invoke, I think it would be helpful to have the name be a bit more descriptive -- something like "dist-overwrite-test-product" or something.
___
*[dister/distertester/distertester.go, line 171 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMavDEBC6YtFq6CWdPR:-MMavDEBC6YtFq6CWdPS:b-jr0v77) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L171)):*
> ```Go
> 			"foo": {
> 				Build: distgoconfig.ToBuildConfig(&distgoconfig.BuildConfig{
> 					MainPkg: stringPtr("foo"),
> ```

Same feedback for more descriptive name
___
*[dister/distertester/distertester.go, line 188 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMavFP33C4hxpqjPpkl:-MMavFP33C4hxpqjPpkm:b-mf564y) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L188)):*
> ```Go
> 	internalFiles := map[string][]byte{
> 		"godel/config/dist-plugin.yml": projectCfgBytes,
> 		"foo/foo.go":                   []byte(`package main; func main() {}`),
> ```

In this case, `main.go` is probably fine
___
*[dister/distertester/distertester.go, line 189 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMavJdc2S7pw_dY7Scv:-MMavJdc2S7pw_dY7Scw:brybzs9) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L189)):*
> ```Go
> 		"godel/config/dist-plugin.yml": projectCfgBytes,
> 		"foo/foo.go":                   []byte(`package main; func main() {}`),
> 		"go.mod": []byte(`module foo
> ```

Would use more descriptive module name

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/191)* status: 0 of 2 files reviewed, 5 unresolved discussions (waiting on @nmiyake)

___
*[changelog/@unreleased/pr-191.v2.yml, line 4 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMb-Ut-3IMkOfswwFCb:-MMq7uP8DHvHEdFXB7z1:b-d8kzr3) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/changelog/@unreleased/pr-191.v2.yml#L4)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Can you make this changelog more precise/descriptive? For example:

```
Adds the distertester.RunDistOverwritesTest function. This function can be used by custom dister implementations to verify that the "dist" operation does not fail when it has to overwrite previous output.
```

I think something like the above makes the change much more legible/understandable when reviewing release notes.
</blockquote></details>

nice, I like that better.
___
*[dister/distertester/distertester.go, line 167 at r1](https://reviewable.io/reviews/palantir/distgo/191#-MMXIjtF9N17WuJ3Avqi-r1-167:-MMq7xX5ACvCJs0yc6mb:b-ax6acm) ([raw file](https://github.com/palantir/distgo/blob/3d760f62566fdb12b04823525b7c565bb638575c/dister/distertester/distertester.go#L167)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Thanks for the explanation! Agree with the sentiment that, since this test is meant to be more targeted, it doesn't need to have as much customizability as the other one -- if every client is going to have to repeat some amount of code to run the test, then centralizing that is a good thing.

However, my suggestion would be to change the provided input from `distType string` to `distgoconfig.DistersConfig` instead -- main thinking here is that a dister may require a certain amount of config, so allowing the config for it to be specified would be helpful.
</blockquote></details>

I agree with that sentiment. updated to take the `distgo.distersConfig` instead
___
*[dister/distertester/distertester.go, line 253 at r1](https://reviewable.io/reviews/palantir/distgo/191#-MMXIjtF9N17WuJ3Avqi-r1-253:-MMq86YbCvv--0ecYDpl:b9zb7zn) ([raw file](https://github.com/palantir/distgo/blob/3d760f62566fdb12b04823525b7c565bb638575c/dister/distertester/distertester.go#L253)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Got it -- maybe it has to do with the timestamps used for caching. Would you mind checking to see if there's a way to trigger a "dist" that always makes it run the build? Or if having a dist that doesn't use build output does so? If we could find that, then we could test that directly rather than using an approach like this.
</blockquote></details>

I was able to repro the issue by running the build task once and then running dist with `--force` to exclude the cached timestamps. I like this solution much better since we strictly test the dist task running twice and this removes the sleep.
___
*[dister/distertester/distertester.go, line 169 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMauxR7201BSX6gf9QS:-MMq84LrAP-FmXG0ZfAL:b-896fix) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L169)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

I know we often use this naming for tests we write, but since this is part of a test API that others will invoke, I think it would be helpful to have the name be a bit more descriptive -- something like "dist-overwrite-test-product" or something.
</blockquote></details>

Done.
___
*[dister/distertester/distertester.go, line 171 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMavDEBC6YtFq6CWdPR:-MMq852VB6FkdAA1qkn3:b-896fix) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L171)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Same feedback for more descriptive name
</blockquote></details>

Done.
___
*[dister/distertester/distertester.go, line 188 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMavFP33C4hxpqjPpkl:-MMq85WDDZIEXEBgMsRv:b-896fix) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L188)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

In this case, `main.go` is probably fine
</blockquote></details>

Done.
___
*[dister/distertester/distertester.go, line 189 at r2](https://reviewable.io/reviews/palantir/distgo/191#-MMavJdc2S7pw_dY7Scv:-MMq85uQ6_XUGcXpzrTn:b-896fix) ([raw file](https://github.com/palantir/distgo/blob/a868a3f603f79b2f8183a42f063aecb8309e2147/dister/distertester/distertester.go#L189)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Would use more descriptive module name
</blockquote></details>

Done.

nmiyake (changes requested):
Thanks for investigating and updating implementation -- I think this is a lot better! Few more points of feedback.

> Reviewed 2 of 2 files at r3.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/191#-:-MMqGKKe-FOAVJzWtmeq:b-xa8at6)* status: all files reviewed, 2 unresolved discussions (waiting on @tabboud)

___
*[dister/distertester/distertester.go, line 253 at r1](https://reviewable.io/reviews/palantir/distgo/191#-MMXIjtF9N17WuJ3Avqi-r1-253:-MMqEgD68p9sGMsbx7xh:b-7yxjt6) ([raw file](https://github.com/palantir/distgo/blob/3d760f62566fdb12b04823525b7c565bb638575c/dister/distertester/distertester.go#L253)):*
<details><summary><i>Previously, tabboud (Tony Abboud) wrote…</i></summary><blockquote>

I was able to repro the issue by running the build task once and then running dist with `--force` to exclude the cached timestamps. I like this solution much better since we strictly test the dist task running twice and this removes the sleep.
</blockquote></details>

Great, thanks! I assume that the "build" task is still necessary because many disters won't actually create an output without binaries, so this makes sense.
___
*[dister/distertester/distertester.go, line 161 at r3](https://reviewable.io/reviews/palantir/distgo/191#-MMqFRFn8MEHEk0Y7Mzu:-MMqFRFn8MEHEk0Y7Mzv:bi9qmh) ([raw file](https://github.com/palantir/distgo/blob/c955a875b6fc1a4e16bf36c43b9cff05733bc577/dister/distertester/distertester.go#L161)):*
> ```Go
> // RunDistOverwritesTest verifies that running the "dist" task multiple times with
> // the provided DistersConfig will succeed without error.
> func RunDistOverwritesTest(t *testing.T,
> ```

Can you update the test name to be something like `RunRepeatedDistTest` or something like that? That naming would be more accurate since, strictly speaking, this function is only testing that forcing the "dist" task to run multiple times with the same build input will not fail (so if a dister implementation just skips outputting or handles this scenario in some other way, this test can still pass).

Would also update the comment slightly to reflect that -- something along the lines that the dist task is run multiple times with the same build input and is run in a way that forces the repeated dist to occur by ignoring the cache.
___
*[dister/distertester/distertester.go, line 167 at r3](https://reviewable.io/reviews/palantir/distgo/191#-MMqELcc-M8jultgUKCW:-MMqELcc-M8jultgUKCX:b9hxgk5) ([raw file](https://github.com/palantir/distgo/blob/c955a875b6fc1a4e16bf36c43b9cff05733bc577/dister/distertester/distertester.go#L167)):*
> ```Go
> ) {
> 	var (
> 		productName = "dist-overwrite-test-product"
> ```

nit: can you make this `const productName = "..."`

tabboud (commented):
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/191)* status: 1 of 2 files reviewed, 1 unresolved discussion (waiting on @nmiyake)

___
*[dister/distertester/distertester.go, line 253 at r1](https://reviewable.io/reviews/palantir/distgo/191#-MMXIjtF9N17WuJ3Avqi-r1-253:-MMqHnnr48zw9iOQwIn2:b-ati1zi) ([raw file](https://github.com/palantir/distgo/blob/3d760f62566fdb12b04823525b7c565bb638575c/dister/distertester/distertester.go#L253)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Great, thanks! I assume that the "build" task is still necessary because many disters won't actually create an output without binaries, so this makes sense.
</blockquote></details>

yep that's exactly right.
___
*[dister/distertester/distertester.go, line 161 at r3](https://reviewable.io/reviews/palantir/distgo/191#-MMqFRFn8MEHEk0Y7Mzu:-MMqJALU8XtjR22YFYNh:b4ztjwd) ([raw file](https://github.com/palantir/distgo/blob/c955a875b6fc1a4e16bf36c43b9cff05733bc577/dister/distertester/distertester.go#L161)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

Can you update the test name to be something like `RunRepeatedDistTest` or something like that? That naming would be more accurate since, strictly speaking, this function is only testing that forcing the "dist" task to run multiple times with the same build input will not fail (so if a dister implementation just skips outputting or handles this scenario in some other way, this test can still pass).

Would also update the comment slightly to reflect that -- something along the lines that the dist task is run multiple times with the same build input and is run in a way that forces the repeated dist to occur by ignoring the cache.
</blockquote></details>

I like that better. Updated the test and doc comment accordingly.
___
*[dister/distertester/distertester.go, line 167 at r3](https://reviewable.io/reviews/palantir/distgo/191#-MMqELcc-M8jultgUKCW:-MMqHZv-6Qwf_yfYaRDH:b-9jiznu) ([raw file](https://github.com/palantir/distgo/blob/c955a875b6fc1a4e16bf36c43b9cff05733bc577/dister/distertester/distertester.go#L167)):*
<details><summary><i>Previously, nmiyake (Nick Miyake) wrote…</i></summary><blockquote>

nit: can you make this `const productName = "..."`
</blockquote></details>

I previously needed to take the address of it for the `MainPkg` struct field, but this ends up defaulting to the repo root, so I no longer need it to be a var (alternatively I had the `stringPtr` func as well). 
Updated it to const instead.

nmiyake (approved):
Looks good to me -- thanks for making the PR and iterating on it!

> Reviewed 1 of 1 files at r4.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/191#-:-MMvSA8mESM1ga8X3G2H:bhzxhyc)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #193, opened 2020-11-25
Add groupID to repeated dist test

tabboud:
## Before this PR
Some internal disters require the publisher groupID in order to build the dist correctly. Our internal dister appears to be the only one that requires extra metadata outside of the `DistersConfig` to build the dist, which makes this feel like a safe change. 

## After this PR

==COMMIT_MSG==
Update configuration used internally by the repeated dist test to specify a publish groupID, as there are some disters that require this information for the dist operation.
==COMMIT_MSG==

## Possible downsides?
There may be other disters, now or in the future, that will require additional metadata, in which case it might make more sense to be less opinionated and allow users to provide the full product config.

cc @nmiyake 

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/193)

### DIFF
--- a/dister/distertester/distertester.go
+++ b/dister/distertester/distertester.go
@@ -178,7 +178,7 @@ func RunRepeatedDistTest(t *testing.T,
 					Disters: distgoconfig.ToDistersConfig(&distersCfg),
 				}),
 				Publish: distgoconfig.ToPublishConfig(&distgoconfig.PublishConfig{
-					GroupID:  stringPtr("com.palantir.test"),
+					GroupID: stringPtr("com.palantir.test"),
 				}),
 			},
 		}),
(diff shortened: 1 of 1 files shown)


tabboud:
I would usually add a `no-changelog` label to this since it's entirely internal, but I do not see a way to do that here.

Also, I have verified this works as intended locally with our internal disters.

nmiyake (approved):
Thanks! Yes, agree that hard-coding this part makes sense in this case, and agree with your assessment that if we see more issues then it would make sense to allow users to specify config (although would likely be by adding API call so that breaking change does not occur).

> Reviewed 1 of 1 files at r2.
> *[Reviewable](https://reviewable.io/reviews/palantir/distgo/193#-:-MN0Hzyc2wv1tNNgDq__:b7wff12)* status: :shipit: complete! all files reviewed, all discussions resolved

PR #194, opened 2020-12-02
Add RunMultipleAssetDockerBuilderTest function

nmiyake:
## Before this PR
The `RunAssetDockerBuilderTest` only accepted a single asset provider, which meant that it was not possible to test Docker assets with non-builtin dist assets.

## After this PR

==COMMIT_MSG==
Adds the RunMultipleAssetDockerBuilderTest function to allow testing a Docker builder asset in conjunction with other assets.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/194)


tabboud (approved):
👍

PR #200, opened 2021-03-27
github publisher: provide ability to add 'v' prefix for releases

### LINKED ISSUE #199: github publisher: add ability to publish with 'v' prefix
## What happened?
Currently, the "github" publisher only supports creating releases that do not have a 'v' prefix. That is, a github publisher run on a project with the version "v1.2.3" and "1.2.3" will both result in the created GitHub release and tag being "1.2.3".

## What did you want to happen?
It should be possible to specify that the release created by the GitHub publisher should have the 'v' prefix. This can be controlled via a flag/configuration (no need to auto-detect whether or not the prefix should exist), but should be made possible.

Having a "v" prefix for releases is common (especially for Go modules projects), so this should be supported by the publisher.

nmiyake:
Fixes #199

## Before this PR
The tags and releases created by the GitHub publisher never have a "v" prefix (even if the project itself has that prefix for a tag/version)

## After this PR

==COMMIT_MSG==
Adds configuration and flag to github publisher that adds a 'v'
prefix to the version used for the tag/release created by the
publisher.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/200)

### DIFF
--- a/changelog/@unreleased/pr-200.v2.yml
+++ b/changelog/@unreleased/pr-200.v2.yml
@@ -1,0 +1,8 @@
+type: improvement
+improvement:
+  description: |-
+    Adds configuration and flag to github publisher that adds a 'v'
+    prefix to the version used for the tag/release created by the
+    publisher.
+  links:
+  - https://github.com/palantir/distgo/pull/200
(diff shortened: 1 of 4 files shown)


tabboud (approved):
👍

PR #222, opened 2021-06-03
Fix: URL encode the Artifactory publisher property fields

tabboud:
## Before this PR
The Artifactory publisher properties were not URL encoded, so any key/values that contain invalid URL parameters would fail to publish.

## After this PR

==COMMIT_MSG==
URL encode the Artifactory publisher property fields

Fixes a bug where the Artifactory properties were inserted into the URL without any encoding which caused failures when publishing. The properties will now be encoded before appending to the Artifactory URL.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/222)

### DIFF
--- a/publisher/common.go
+++ b/publisher/common.go
@@ -202,7 +202,7 @@ func (b *BasicConnectionInfo) UploadFile(fileInfo FileInfo, baseURL, artifactNam
 		uploadMsgParts = append(uploadMsgParts, filePath)
 	}
 	uploadMsgParts = append(uploadMsgParts, "to", rawUploadURL)
-	distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf(strings.Join(uploadMsgParts, " ")), dryRun)
+	distgo.PrintlnOrDryRunPrintln(stdout, strings.Join(uploadMsgParts, " "), dryRun)
 
 	if !dryRun {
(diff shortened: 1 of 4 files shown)


publisher/common.go
```diff
@@ -202,7 +202,7 @@ func (b *BasicConnectionInfo) UploadFile(fileInfo FileInfo, baseURL, artifactNam
 		uploadMsgParts = append(uploadMsgParts, filePath)
 	}
 	uploadMsgParts = append(uploadMsgParts, "to", rawUploadURL)
-	distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf(strings.Join(uploadMsgParts, " ")), dryRun)
```
tabboud, line 205: This `fmt.Sprintf` is not actually needed and in fact caused issues when printing any URLs contained in the `uploadMsgParts` since the URL can now contain encoded fields (i.e. `%2F`) which caused the output to be something like `%!(MISSING)` since it interprets the `%` as a variable replacement.

nmiyake: Yes, this makes sense -- thanks!

nmiyake (approved):
Thanks! PR looks great -- appreciate the thorough documentation/explanation and including the test case.

PR #249, opened 2022-02-24
Add "filter-artifact-names" and "exclude-artifact-names" to publishers

nmiyake:
## Before this PR
The built-in publishers would always attempt to publish all artifacts for a given distribution.

Generally, this is the desired behavior. However, if there are disters that create multiple different kinds of dist outputs, then there may be scenarios in which an end user would want to publish certain outputs using a particular publisher and other outputs using another publisher, and this was not previously possible.

Another scenario where this is useful is when a publish only partially succeeds and the user wants to perform the operation again but only for the artifacts that were not published.

## After this PR

==COMMIT_MSG==
Adds flags to the built-in publishers that enables controlling the behavior
of the artifacts that will be published by the operation.

If "--filter-artifact-names" is specified, then only artifacts with names
that match the specified regular expression are published.

If "--exclude-artifact-names" is specified, then any artifacts with names
that match the specified regular expression are not published.
==COMMIT_MSG==

## Possible downsides?
The primary downside/risk is that this allows for more complexity in usage. However, I believe that the benefit gained from the new workflows that this allows in specialized cases outweighs this downside.

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/249)

### DIFF
--- a/changelog/@unreleased/pr-249.v2.yml
+++ b/changelog/@unreleased/pr-249.v2.yml
@@ -1,0 +1,13 @@
+type: feature
+feature:
+  description: |-
+    Adds flags to the built-in publishers that enables controlling the behavior
+    of the artifacts that will be published by the operation.
+
+    If "--filter-artifact-names" is specified, then only artifacts with names
+    that match the specified regular expression are published.
+
+    If "--exclude-artifact-names" is specified, then any artifacts with names
+    that match the specified regular expression are not published.
+  links:
+  - https://github.com/palantir/distgo/pull/249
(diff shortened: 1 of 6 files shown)


tabboud (approved):
I am onboard with this approach. It should solve some issues we had internally around conditional publishing as well which is great. I had previously explored this feature using declarative config rather than CLI flags ([link](https://github.com/palantir/distgo/pull/188/files)) but I actually prefer this approach since it doesn't add any more configuration but still allows a similar behavior.

PR #253, opened 2022-03-05
Artifactory publisher: only upload POM if at least 1 artifact is uploaded

nmiyake:
## Before this PR
The artifactory publish operation would create and upload a POM file even if no artifacts were uploaded (for example, because of filters applied to artifact names).

## After this PR

==COMMIT_MSG==
Updates Artifactory publisher to upload a POM file only if the publish operation publishes at least 1 artifact.
==COMMIT_MSG==

## Possible downsides?

---
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/253)

### DIFF
--- a/changelog/@unreleased/pr-253.v2.yml
+++ b/changelog/@unreleased/pr-253.v2.yml
@@ -1,0 +1,6 @@
+type: fix
+fix:
+  description: Updates Artifactory publisher to upload a POM file only if the publish
+    operation publishes at least 1 artifact.
+  links:
+  - https://github.com/palantir/distgo/pull/253
--- a/publisher/artifactory/integration_test/integration_test.go
+++ b/publisher/artifactory/integration_test/integration_test.go
@@ -129,6 +129,47 @@ products:
 				},
 			},
 			{
+				Name: "skips POM publish if no artifacts are uploaded",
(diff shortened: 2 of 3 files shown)


nmiyake:
@tabboud this is the PR that deals with the POM issue I mentioned to you earlier

tabboud (approved):
Sorry about the delay.

The logic here seems sane to me. The comment is also helpful since we're now proxying the artifact info through what was uploaded (or lack there of).

Slightly tangential question:
For the filtering logic, I noticed we only update the `DistInfos` map (see below). Is there a reason we don't also update the `DistOutputInfos.DistIds` to include/exclude the DistIds altogether? This works fine in this PR since lower level logic handles the filtering, but just something I noticed.
https://github.com/palantir/distgo/blob/8571d4440fd665b11eeec4df3eab2b1f598e61ff/publisher/common.go#L358-L359

nmiyake:
Thanks for the review!

Yes, I did consider doing a post-filter to "re-normalize" the struct (for example, remove dist IDs for dists that no longer had artifacts after filtering), but thought that conceptually it might be cleaner for the filter logic to just filter the artifacts. However, this is definitely fuzzy -- I didn't do a full analysis as to what the difference would be between doing that or not doing that.

PR #307, opened 2023-01-06
Add docker buildx builder

andybradshaw:
## After this PR

==COMMIT_MSG==
Add support for multi-architecture builds with a docker `buildx` builder 
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/307)


andybradshaw:
Not entirely sure we should still be splitting this across the various repos/plugin assets, but it basically mirrors the existing dockerbuilder, just with a different underlying command structure, so figured it was okay to add here.

dockerbuilder/buildxdockerbuilder/dockerbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+	cmd := exec.Command("docker", "buildx", "inspect")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return err
+	}
+
+	if bytes.Contains(out, []byte("docker-container")) {
+		return nil
+	}
+	var driverOptArgs []string
```
nmiyake, line 93: nit, but since `driverOptArgs` are appended to `args`, I think it's more readable to have this block after the declaration of the `args` variable

dockerbuilder/buildxdockerbuilder/dockerbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// ensureDockerContainerDriver ensures there is a buildx builder that uses the docker-container driver, currently
+// required for multi-arch support, until docker finishes supporting multi-arch containers in the daemon. If one does
+// not exist, create one and set it to the default.
+func (b *BuildxDockerBuilder) ensureDockerContainerDriver() error {
+	cmd := exec.Command("docker", "buildx", "inspect")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return err
```
nmiyake, line 87: Helpful to include `string(out)` as part of the error for debugging

dockerbuilder/buildxdockerbuilder/dockerbuilder.go
```diff
@@ -0,0 +1,111 @@
+// Copyright 2016 Palantir Technologies, Inc.
```
nmiyake, line 1: nit, but since this is a new file, should be 2023 (I believe the license plugin on the repo would properly add the current year, but assume this is from a copy-paste of the default builder)

dockerbuilder/buildxdockerbuilder/dockerbuilder.go
```diff
@@ -0,0 +1,111 @@
...
+
+func (b *BuildxDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, verbose, dryRun bool, stdout io.Writer) error {
+	// if we actually need to build the image, we need to execute against a working buildx environment
+	if !dryRun {
+		if err := b.ensureDockerContainerDriver(); err != nil {
+			return err
+		}
+	}
+
+	dockerBuilderOutputInfo := productTaskOutputInfo.Product.DockerOutputInfos.DockerBuilderOutputInfos[dockerID]
```
nmiyake, line 54: It looks like most of this logic is shared with the default Docker builder.

Zooming out, it looks like the main differences between the `buildx` builder and the default builder are:
* Presence of `DriverOpts` in configuration
* Validating presence of Docker container driver
* Running `buildx build` instead of `build` as the base command

Given all of this, I think it may make more sense to implement the `buildx` functionality as a new optional behavior on the default builder rather than as an entirely new builder. That way, we wouldn't need to duplicate the core logic or the scaffolding for things like configuration, integration tests, etc.

We could either have a boolean flag or a default-empty optional configuration block and have either of those things signal that `buildx` should be used, in which case the `buildx` argument would be added to the `args` slice and the logic that ensures the driver would be run.

nmiyake (changes requested):
Thanks for the PR! I'm supportive of adding the functionality, but per the feedback, would suggest that we add the functionality by adding a configuration point/value on the default Docker builder rather than creating an entirely new builder.

I would also like there to be unit and integration tests added to where they already exist for Docker builders to verify that the functionality works in the desired manner.

changelog/@unreleased/pr-307.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: improvement
+improvement:
+  description: Add support for multi-architecture builds with a docker `buildx` builder
```
nmiyake, line 3: The content here will likely change if the PR feedback is followed, but ultimately this will serve as the release note/documentation, so when the final approach is determined it would be nice to expand this to describe more of what changed and how it can be used.

nmiyake:
I've updated the RFC with this feedback, but from a big picture perspective I think this PR should be changed to:

1. Define the concept/contract for where the output for where the `docker buildx` manifest directory and output should go
2. Make it so the `buildx` build operation writes its output in a standardized manner to the location decided above (rather than delegating that to the managed builder)
3. Either have `buildx` as its own separate builder (as this PR does) OR modify the default builder so that it supports writing OCI output and have a mode flag that controls whether the builder builds just daemon images, just OCI image output, or daemon image + OCI output
  * If it's the former, implementation should still be refactored so that default and dockerx builder mostly share the underlying implementation rather than duplicating the code

andybradshaw:
> Would suggest that we add the functionality by adding a configuration point/value on the default Docker builder rather than creating an entirely new builder.

Seems like there's not a good way right now to add any "options" to the builder, so we probably need to add a different constructor which gives us that flexibility. Added something with `Options` in the latest commit, but happy to take a different approach. Seems like everything we provide to the builder is optional at the moment, although this might not be the case with the Buildx output options.

> Define the concept/contract for where the output for where the `docker buildx` manifest directory and output should go

This makes sense, and will make it more obvious/consistent between `build` and `push` tasks. I've done this by adding a `ProductDockerDistOutputDir` which takes a DockerID and produces a path in the dist dir. I first attempted to handle this by adding something to the `DockerBuilderOutputInfo` struct, but that seemed to not fit very well. Let me know if you had other ideas about how to handle this.

> Make it so the `buildx` build operation writes its output in a standardized manner to the location decided above (rather than delegating that to the managed builder)

I have some concerns around how this can interact with the existing implementation around "buildArgs"... right now, the builder has very few opinions about what it's doing, but adding this behavior here means not all provided args will be respected (or rather, it's up to how docker parses them). I think it's more correct to have this behavior here, but think it means we might need to scope down what consumers can provide.

andybradshaw:
Another annoying thing we're probably going to need to handle here is the fact that the `oci` artifact isn't quite what we want. In the resulting tarball we produce, the `index.json` is an image index which contains all the tags that we applied during the build:

```json
{
  "schemaVersion": 2,
  "manifests": [
    {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:be0dd4591f7ceba283693862ee1d1cdc51cda4c32631bfd520ebba3a4810767b",
      "size": 705,
      "annotations": {
        "io.containerd.image.name": "palantir/test:0.0.1",
        "org.opencontainers.image.created": "2023-01-05T16:49:10Z",
        "org.opencontainers.image.ref.name": "0.0.1"
      }
    },
    {
      "mediaType": "application/vnd.oci.image.index.v1+json",
      "digest": "sha256:be0dd4591f7ceba283693862ee1d1cdc51cda4c32631bfd520ebba3a4810767b",
      "size": 705,
      "annotations": {
        "io.containerd.image.name": "palantirtest:snapshot",
        "org.opencontainers.image.created": "2023-01-05T16:49:10Z",
        "org.opencontainers.image.ref.name": "snapshot"
      }
    }
  ]
}
```

I don't think this is actually what we want as I've run into issues trying to push this manifest to a registry, since it sort of breaks how these things are meant to be used (What is the named reference to the above image index? Do both `0.0.1` and `snapshot` point to it? If so, why duplicate this information?). I think we need to follow the first digest and use that image index as the top-level image-index instead (which points to the various architectures in the index), which we'll push for each of the tags:

```json
{
   "mediaType": "application/vnd.oci.image.index.v1+json",
   "schemaVersion": 2,
   "manifests": [
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "digest": "sha256:ae76b86cdfca59a11fc30d0986ef46742da678d379c9d291ccc1eb8401a4c420",
         "size": 1115,
         "platform": {
            "architecture": "amd64",
            "os": "linux"
         }
      },
      {
         "mediaType": "application/vnd.oci.image.manifest.v1+json",
         "digest": "sha256:9e7c6cea504c77e1d56ec9fa33a0bf6bb79cd2372c20c866eed2ba6c1f672430",
         "size": 1115,
         "platform": {
            "architecture": "arm64",
            "os": "linux"
         }
      }
   ]
}
```

It's a little annoying that we have to do this at all, but with the "output" responsibilities shifting from the managed plugin to this repo, I think we'll need to move that here 🙁

nmiyake:
I don't fully understand the last comment... I'm not as familiar with the index files, but from the examples you posted above, in the first one it looks like the index file itself points to different index files (referenced by tag name?), while the second one references the specific manifests of the images.

It might help to make the expected outputs for each case more explicit.

## "build" step

### For regular/current Docker builder:
* Result of "docker build" is that image is built and pushed/published in local Docker daemon using all of the associated tags

### Proposal for Docker buildx builder:
* Result of "docker buildx" is:
  * A TGZ file that contains the "full" Docker buildx output: a top-level image index file and all of the individual images that are referenced from that file laid out in the proper format (where just distributing/publishing the TGZ is sufficient)
  * Tags are *not* part of this output -- they are only used/referenced on push

## "push" step

### For regular/current Docker builder:
* "Push" operation issues a "push" command to local Docker daemon using tags for the build and the provided publish URL
* This depends on "build" having already been run and tagging the images in the Docker daemon with the expected tags/names

### Proposal for Docker buildx builder:
* "Push" operation pushes the TGZ (index file + individual images) to the publish URL for each of the tags
* Depends on the TGZ output already being created/existing in expected location, and the push to remote is the first/only thing that actually applies the tags

Is the above correct and what we want? If it's off, can you identify what's off about it/clarify what the proposal actually is?

andybradshaw:
>  A TGZ file that contains the "full" Docker buildx output: a top-level image index file and all of the individual images that are referenced from that file laid out in the proper format (where just distributing/publishing the TGZ is sufficient)
> 
>  Tags are _not_ part of this output -- they are only used/referenced on push

Yes, I believe this is the output we want, which means we need to modify the tarball produced by the build to get rid to the level of indirection that the `buildx` process is currently introducing.

> the push to remote is the first/only thing that actually applies the tags

Yeah, that seems right. Tagging during the build isn't really needed, since the artifact will be the same and we know the full set of tags we want to push during that task.

andybradshaw:
@nmiyake think this is probably ready for another pass if we're in agreement about the oci-layout as the "dist" that we want here

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +88,123 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+}
+
+// ensureDockerContainerDriver ensures there is a buildx builder that uses the docker-container driver, currently
+// required for multi-arch support, until docker finishes supporting multi-arch containers in the daemon. If one does
+// not exist, create one and set it to the default.
+func (d *DefaultDockerBuilder) ensureDockerContainerDriver(verbose, dryRun bool, stdout io.Writer) error {
+	cmd := exec.Command("docker", "buildx", "inspect")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return err
```
nmiyake, line 153: Likely helpful to include `string(out)` as a wrapped message for debugging

nmiyake (changes requested):
Thanks, I think this approach is an improvement. Given that this task isn't technically a `dist` task, having the "dist" output be the directory that uses an OCI layout is likely fine. There's a part of me that wonders whether it would be better to produce a TGZ of the OCI layout so we still have a single-file dist output, but there may not really be a practical reason to do this if the only consumer is the `publish` task and that task wants/needs an on-disk directory structure anyway.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +88,123 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
 		}
 		args = append(args, buildArgsFromScript...)
 	}
-	args = append(args, contextDirPath)
 
-	cmd := exec.Command("docker", args...)
-	return distgo.RunCommandWithVerboseOption(cmd, verbose, dryRun, stdout)
+	if d.Output&OCITarball != 0 {
```
nmiyake, line 92: As discussed in sync, add check here and error if `d.Output == 0` and/or if none of the relevant bits are set (we expect that every `build` operation should build at least 1 valid output).

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +92,129 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
 		}
 		args = append(args, buildArgsFromScript...)
 	}
-	args = append(args, contextDirPath)
 
-	cmd := exec.Command("docker", args...)
-	return distgo.RunCommandWithVerboseOption(cmd, verbose, dryRun, stdout)
+	if d.OutputType == 0 {
+		return errors.New("Output type of docker builder must be specified")
```
nmiyake, line 97: We should keep this, but I don't think it catches all cases -- if someone provides `1 << 3` right now, I believe that it won't be 0, but also won't match either of the existing masks.

We should probably have a `buildRan := false` boolean, flip it to true for any mask match, and then throw an error at the end of all flag checks if it is still `false`

changelog/@unreleased/pr-307.v2.yml
nmiyake: Changelog should be updated. It should reflect that the new functionality has been added, but should also mention that it is only exposed at an API level and that this change alone does not make it possible for end users of `distgo` to create multi-architecture Docker images.

nmiyake (changes requested):

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
andybradshaw: Added a const for all the build types and checked against that

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,127 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+		return err
+	}
+	idxManifest, err := index.IndexManifest()
+	if err != nil {
+		return err
+	}
+	if len(idxManifest.Manifests) == 0 {
+		return errors.New("Top-level OCI image index does not contain any manifests. While this is a valid image index, it is unexpected and likely means something erroneous happened earlier in the build")
+	}
+	if err := os.Rename(path.Join(destOCILayoutDir, "blobs", idxManifest.Manifests[0].Digest.Algorithm, idxManifest.Manifests[0].Digest.Hex), path.Join(destOCILayoutDir, "index.json")); err != nil {
```
nmiyake, line 150: I believe this should be `filepath.Join` since it's for dealing with renames/moves on the local disk.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,127 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+		return err
+	}
+	return nil
+}
+
+// ensureDockerContainerDriver ensures there is a buildx builder that uses the docker-container driver, currently
```
nmiyake, line 156: Reword to make a bit more natural/readable:

> ensureDockerContainerDriver ensures there is a buildx builder that uses the docker-container driver, which is required for building multi-arch images.  If a buildx builder does not exist, creates one and sets it as the default. This is required until docker finishes supporting multi-arch containers in the daemon.

Also, it would be helpful if there was a linked issue for this (if one exists) and an explanation of what would be changed once this support does exist -- as it stands right now, I'm not sure what the criteria is for this being removed/changed when the support is eventually added/complete.

distgo/product.go
```diff
@@ -99,6 +100,10 @@ func (p *ProductTaskOutputInfo) ProductDockerDistArtifactPaths() map[DockerID]ma
 	return ProductDockerDistArtifactPaths(p.Project, p.Product, p.Deps)
 }
 
+func (p *ProductTaskOutputInfo) ProductDockerOCIDistOutputDir(dockerID DockerID) string {
```
nmiyake, line 103: Please add documentation -- I think something like the following would work:

```
// ProductDockerOCIDistOutputDir returns the output directory for the Docker OCI dist outputs, which is
// "{{ProjectDir}}/{{OutputDir}}/{{ProductID}}/oci-{{DockerID}}". If the builder for a given DockerID uses the buildx builder for multi-architecture images, the output is written to this directory.
//
// Note that this scheme uses the namespace for dist outputs, so if a product has a DockerID "X" and a dist with
// DistID "oci-X", then the output directories will be the same and the behavior will be undefined -- this is a
// known issue/risk that we are accepting as part of the design.
```

nmiyake: Minor, but I would move this function next to `ProductDistOutputDir`

distgo/product.go
```diff
@@ -99,6 +100,10 @@ func (p *ProductTaskOutputInfo) ProductDockerDistArtifactPaths() map[DockerID]ma
...
+	return ProductDistOutputDir(p.Project, p.Product, DistID(fmt.Sprintf("oci-%s", dockerID)))
```
nmiyake, line 104: I think the path here should mirror the other outputs by including the `{{Version}}` as part of the path after `Product` -- that way if the Docker builder is run on multiple versions of the product the output is kept separate.

Also, my example comment above about conflicting with `dist` may not completely be true given how we construct this output path.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,127 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+
+	if bytes.Contains(out, []byte("docker-container")) {
+		return nil
+	}
+	var driverOptArgs []string
+	for _, opt := range d.BuildxDriverOpts {
+		driverOptArgs = append(driverOptArgs, "--driver-opt", opt)
+	}
+	createContextArgs := []string{"context", "create", string(dockerID)}
```
nmiyake, line 173: Can you add comment here with link to the docs you included in the Slack chat to document that we're intentionally doing this as a 2 step process because of the known issue?

nmiyake (changes requested):
Thanks! Some minor requests for modification.

Also, can you add at least one test that invokes this behavior in `dockerbuilder/defaultdockerbuilder/integration_test/integration_test.go`?

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,127 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	return nil
+}
+
+// extractToOCILayout is responsible for converting the buildx OCI tarball output to a compatible OCI layout on disk.
+// The buildx tarball adds a layer of indirection which doesn't seem to play nicely with some registries; the top-level
+// image index produced contains a manifest per-tag, which point to the "actual" image index we want to publish. Since
+// we know all the rendered tags at publish time, we can move the "actual" image index back to the top level and do a
+// publish per-tag.
+func (d *DefaultDockerBuilder) extractToOCILayout(destOCILayoutDir, sourceOCITarball string) error {
+	if err := archiver.DefaultTar.Unarchive(destOCILayoutDir, sourceOCITarball); err != nil {
```
gcampbell12, line 136: Looks like this is the wrong way round, should be source tarball and the directory to extract it into
```suggestion
	if err := archiver.DefaultTar.Unarchive(sourceOCITarball, destOCILayoutDir); err != nil {
```

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,127 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
-	return distgo.RunCommandWithVerboseOption(cmd, verbose, dryRun, stdout)
+	if d.OutputType&allOutputs == 0 {
+		return errors.New("A valid output type of docker builder must be specified")
+	}
+
+	if d.OutputType&OCILayout != 0 {
+		if err := d.ensureDockerContainerDriver(dockerID, verbose, dryRun, stdout); err != nil {
+			return err
+		}
+		destDir := productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)
```
gcampbell12, line 106: This directory needs to exist prior to passing it to the build command, could we run it through `os.MkdirAll`?

nmiyake:
Never mind regarding integration test -- integration tests run based on rendered config, and since there's no way to instantiate this builder via config in this PR, it's not possible to test via integration tests.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
andybradshaw: Added.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
andybradshaw: Good catch! Updated

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
andybradshaw: Added a comment and linked to the circle article.

distgo/product.go
andybradshaw: Yep, since this uses `ProductDistOutputDir`, it should pull out the version internally:
```
path.Join(projectInfo.ProjectDir, productOutputInfo.DistOutputInfos.DistOutputDir, string(productOutputInfo.ID), projectInfo.Version, string(distID))
```

distgo/product.go
andybradshaw: Updated and moved.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
andybradshaw: Updated the comment and linked to the docs where it mentions that the docker daemon driver does not support it. I was unable to find a high-level [tracking issue here](https://github.com/docker/roadmap), might be worth creating one as it's pretty unclear the direction the project is taking.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
andybradshaw: 👍 good flag, updated the other locations where we did this as well (context dir and dockerfile path)

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
 		}
 		args = append(args, buildArgsFromScript...)
 	}
-	args = append(args, contextDirPath)
 
-	cmd := exec.Command("docker", args...)
-	return distgo.RunCommandWithVerboseOption(cmd, verbose, dryRun, stdout)
+	if d.OutputType&allOutputs == 0 {
+		return errors.New("A valid output type of docker builder must be specified")
```
nmiyake, line 99: Errors should not start with a capital letter -- that's generally a Go convention because errors are often chained in output.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+
+	if d.OutputType&OCILayout != 0 {
+		if err := d.ensureDockerContainerDriver(dockerID, verbose, dryRun, stdout); err != nil {
+			return err
+		}
+		destDir := productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)
+		if err := os.MkdirAll(destDir, 0755); err != nil {
+			return err
```
nmiyake, line 108: Wrap with an `errors.Wrapf` call so that stack trace is included

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+		}
+		destFile := fmt.Sprintf("%s/image.tar", destDir)
```
nmiyake, line 110: This should be a `filepath.Join(destDir, "image.tar")`

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+}
+
+// extractToOCILayout is responsible for converting the buildx OCI tarball output to a compatible OCI layout on disk.
+// The buildx tarball adds a layer of indirection which doesn't seem to play nicely with some registries; the top-level
+// image index produced contains a manifest per-tag, which point to the "actual" image index we want to publish. Since
+// we know all the rendered tags at publish time, we can move the "actual" image index back to the top level and do a
+// publish per-tag.
+func (d *DefaultDockerBuilder) extractToOCILayout(destOCILayoutDir, sourceOCITarball string) error {
+	if err := archiver.DefaultTar.Unarchive(sourceOCITarball, destOCILayoutDir); err != nil {
+		return err
```
nmiyake, line 140: Wrap error

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+	index, err := layout.ImageIndexFromPath(destOCILayoutDir)
+	if err != nil {
+		return err
```
nmiyake, line 144: Wrap error

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+	idxManifest, err := index.IndexManifest()
+	if err != nil {
+		return err
```
nmiyake, line 148: Wrap error

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+	if len(idxManifest.Manifests) == 0 {
+		return errors.New("Top-level OCI image index does not contain any manifests. While this is a valid image index, it is unexpected and likely means something erroneous happened earlier in the build")
```
nmiyake, line 151: Use lowercase first letter and try to maintain as single segment:

```
top-level OCI image index does not contain any manifests -- while this is a valid image index, it is unexpected and likely means something erroneous happened earlier in the build
```

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+	if err := os.Rename(filepath.Join(destOCILayoutDir, "blobs", idxManifest.Manifests[0].Digest.Algorithm, idxManifest.Manifests[0].Digest.Hex), filepath.Join(destOCILayoutDir, "index.json")); err != nil {
+		return err
```
nmiyake, line 154: Wrap error

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+	}
+	return nil
+}
+
+// ensureDockerContainerDriver ensures there is a buildx builder that uses the docker-container driver, currently
+// required for multi-arch support, until docker finishes supporting multi-arch containers in the daemon. If one does
+// not exist, create one and set it to the default.
+
+// ensureDockerContainerDriver ensures there is a buildx builder that uses the docker-container driver, which is
```
nmiyake, line 163: Looks like this comment is now repeated -- delete the section before this one.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+// required for building multi-arch images. If a buildx builder does not exist, creates one and sets it as the default.
+// This is required until docker finishes supporting multi-arch containers in the daemon.
+// https://docs.docker.com/engine/reference/commandline/buildx_create/#driver
+func (d *DefaultDockerBuilder) ensureDockerContainerDriver(dockerID distgo.DockerID, verbose, dryRun bool, stdout io.Writer) error {
+	cmd := exec.Command("docker", "buildx", "inspect")
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("failed to check for existence of buildx drivers: %s", string(out))
```
nmiyake, line 171: Use `errors.Wrapf` to include stack trace and also to include output of the original error (in addition to the string output)

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -60,8 +94,139 @@ func (d *DefaultDockerBuilder) RunDockerBuild(dockerID distgo.DockerID, productT
...
+		driverOptArgs = append(driverOptArgs, "--driver-opt", opt)
+	}
+	// Some CI environments have compatibility issues running with the TLS data in the default context. Creating a new
+	// named context copies the TLS data correctly, and allows for a buildx builder to be created.
+	// https://support.circleci.com/hc/en-us/articles/360058095471-How-To-Use-Docker-Buildx-in-Rem have compatibility issues running with the TLS data in the default context. Creating a new
+	// named context copies the TLS data correctly, and allows for a buildx builder to be created.ote-Docker-
+	createContextArgs := []string{"context", "create", string(dockerID)}
+	createContextCmd := exec.Command("docker", createContextArgs...)
+	if err := distgo.RunCommandWithVerboseOption(createContextCmd, verbose, dryRun, stdout); err != nil {
+		return err
```
nmiyake, line 188: Just noting that these calls are fine as-is (don't need to be wrapped) since they call into other `distgo` functions, and the expectation is that in-library errors will already be the proper type (or wrapped if the underlying cause is due to external libraries)

nmiyake (changes requested):
Thanks! A few minor things to touch up, but should be ready after that.

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
nmiyake: Still needs to be wrapped, but low-pri/will not block.

nmiyake (approved):

PR #309, opened 2023-01-06
Attempt to publish OCI artifact if it exists

andybradshaw:
## After this PR

==COMMIT_MSG==
Attempt to publish OCI artifact if it exists
==COMMIT_MSG==

## Possible downsides?

Biggest downside is adding a fair sized dependency of go-containerregistry to handle the push operations, but we already use this library in quite a few other places when interacting with images/registries and haven't had any issues with it. Alternative approaches would require using an external tool like `skopeo` or `cxtool`, but that would require more coordination with changes to what's included in CI... I think it's preferable to have a slightly larger binary and just do the push ourselves.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/309)


distgo/docker/docker_push.go
```diff
@@ -79,6 +84,41 @@ func runSingleDockerPush(
 	stdout io.Writer) (rErr error) {
 
 	distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("Running Docker push for configuration %s of product %s...", dockerID, productID), dryRun)
+
+	// if an OCI artifact exists, push that. Otherwise, default to pushing the artifact in the docker daemon
+	if _, err := os.Stat(fmt.Sprintf("%s/%s.tar", productTaskOutputInfo.ProductDistOutputDir(distgo.DistID("oci")), dockerID)); err == nil {
```
andybradshaw, line 89: Not sure if this is actually the right thing to test for existence, as `layout.ImageIndexFromPath` eventually checks for an `index.json` file. The tar artifact is the direct output of the buildx process (`--output=oci`), while the index file exists after extracting it. Think it's probably more correct/less confusing to check for the index and have the build process clean up the tarball.

changelog/@unreleased/pr-309.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: feature
+feature:
+  description: Attempt to publish OCI artifact if it exists
```
nmiyake, line 3: This can wait until iteration on the content is finished, but the description will serve as the primary release note/documentation, so I think this should be expanded further -- it should state that it's for the "docker push" operation and describe the new behavior and the situations in which it is triggered (at the detail of a user perspective)

distgo/docker/docker_push.go
```diff
@@ -79,6 +83,41 @@ func runSingleDockerPush(
 	stdout io.Writer) (rErr error) {
 
 	distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("Running Docker push for configuration %s of product %s...", dockerID, productID), dryRun)
+
+	// if an OCI artifact exists, push that. Otherwise, default to pushing the artifact in the docker daemon
+	if _, err := layout.FromPath(productTaskOutputInfo.ProductDistOutputDir(distgo.DistID("oci"))); err == nil {
```
nmiyake, line 88: What's the significance of the `distgo.DistID("oci")` value here? From reading the code, it looks like this is assuming a convention that, for OCI images, the manifest file must be written to the directory that would be use for a dist named `"oci"`.

I don't think this was mentioned in the spec and we may need to think through this further -- I believe this would cause an issue if there were a dist that was actually named "oci", and the way this is set up makes it look like we wouldn't be able to support multiple OCI Docker builder configurations for a single product (since they would all use the same "oci" dist directory).

I think it might make more sense to have this location be based on the DockerID -- something like `"{DockerID}-oci"`. Even if it still models this as a dist directory, I think this approach would reduce the risk of name collisions and also allow for multiple Docker builder configurations.

distgo/docker/docker_push.go
```diff
@@ -79,6 +83,41 @@ func runSingleDockerPush(
...
+		return runOCIPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
+	}
+	return runDockerDaemonPush(dockerID, productTaskOutputInfo, dryRun, stdout)
+}
+
+func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, stdout io.Writer) error {
+	if dryRun {
+		distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("Running Docker OCI push for configuration %s of product %s...", dockerID, productID), dryRun)
```
nmiyake, line 96: This seems to mirror the output of line 85, so I think the behaviors should be equivalent -- namely, the `PrintlnOrDryRunPrintln` call at line 85 should probably be moved into the `runDockerDaemonPush` function, and the `if dryRun` conditional here should be removed so that the output will printed both in dry run and non-dry run modes (the print function will print the `"[DRY RUN]"` prefix when it is a dry run)

nmiyake (changes requested):
Main feedback is around clarifying and updating convention for where the command looks for the OCI index file/manifest.

Once the implementation is settled, test cases should be added to `docker_push_test.go`.

nmiyake:
Provided feedback in RFC as well, but I think that location for OCI manifest should be standardized and made part of `distgo` and then the implementation here should make use of that

distgo/docker/docker_push.go
```diff
@@ -78,6 +82,41 @@ func runSingleDockerPush(
...
+	if _, err := layout.FromPath(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)); err == nil {
+		return runOCIPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
+	}
+	return runDockerDaemonPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
+}
+
+func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, stdout io.Writer) error {
+	distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("Running Docker OCI push for configuration %s of product %s...", dockerID, productID), dryRun)
+	if dryRun {
+		return nil
```
nmiyake, line 95: I think it would be better to more closely mirror the behavior of the `runDockerDaemonPush` function, which will output more information for dry runs.

In this case, I think the dry run check should just wrap/skip the `remote.WriteIndex` call and add output of the general form `distgo.PrintlnOrDryrunPrintln("Writing index for tag %s", tag)` or something like that -- basically, if running in dry-run mode, it would be nice if the output indicated the tags that would have been attempted to be published.

I'm not sure whether the other command only outputs the command run in non-dry-run mode when verbose mode is on, but mirroring that would also help so that looking at output makes it clearer what was published (and if there is an error it's clearer which publish operation failed).

nmiyake (changes requested):

distgo/docker/docker_push.go
```diff
@@ -78,6 +82,40 @@ func runSingleDockerPush(
...
+	if _, err := layout.FromPath(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)); err == nil {
+		return runOCIPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
+	}
+	return runDockerDaemonPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
+}
+
+func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, stdout io.Writer) error {
+	index, err := layout.ImageIndexFromPath(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID))
+	if err != nil {
+		return err
```
nmiyake, line 95: Wrap error

distgo/docker/docker_push.go
```diff
@@ -78,6 +82,40 @@ func runSingleDockerPush(
...
+	}
+	for _, tag := range productTaskOutputInfo.Product.DockerOutputInfos.DockerBuilderOutputInfos[dockerID].RenderedTags {
+		distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("Writing index for tag %s of configuration %s of product %s...", tag, dockerID, productID), dryRun)
```
nmiyake, line 98: Can you update message to "Docker configuration" to make it clearer that this refers to the DockerID?

distgo/docker/docker_push.go
```diff
@@ -78,6 +82,40 @@ func runSingleDockerPush(
...
+		if !dryRun {
+			ref, err := name.ParseReference(tag)
```
nmiyake, line 100: Let's move this logic for `ParseReference` outside of the `dryRun` check -- it looks like all of the logic of this function is local, and if the tag is a format that would fail parsing, it would be good to surface that during the dry run.

distgo/docker/docker_push.go
```diff
@@ -78,6 +82,40 @@ func runSingleDockerPush(
...
+			if err != nil {
+				return err
```
nmiyake, line 102: Wrap error

distgo/docker/docker_push.go
```diff
@@ -78,6 +82,40 @@ func runSingleDockerPush(
...
+			}
+			if err := remote.WriteIndex(ref, index, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
+				return err
```
nmiyake, line 105: Wrap error

nmiyake (changes requested):
Generally looks good -- some minor feedback

nmiyake (approved):

PR #339, opened 2023-03-06
Publish OCI artifacts

andybradshaw:
## After this PR

==COMMIT_MSG==
Add support for publishing OCI artifacts if they exist, and handles both image index and image manifest types.
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/339)


changelog/@unreleased/pr-339.v2.yml
```diff
@@ -0,0 +1,13 @@
+type: fix
+fix:
```
nmiyake, line 2: nit, but this would probably be a "feature" or "improvement" (if this functionality already existed and you were fixing the failure that would be more of a fix, but we rolled back that behavior entirely)

distgo/docker/docker_push.go
```diff
@@ -78,6 +87,104 @@ func runSingleDockerPush(
...
+		// 3. image index (without multiple platforms) pointing to manifests -> read image from index and push manifest
+		// 4. manifest -> re-extract image from tarball and push manifest :(
+		idxManifest, err := index.IndexManifest()
+		if err != nil {
+			return errors.Wrap(err, "failed to read index manifest")
+		}
+		switch idxManifest.MediaType {
+		case types.OCIImageIndex:
+			// check the type of the inner manifests
+			switch idxManifest.Manifests[0].MediaType {
```
nmiyake, line 121: Is `idxManifest.Manifests` guaranteed to be exactly size 1? This will crash if `Manifests` is empty, and if there are more than 1 then they aren't considered?

nmiyake: Also the nested `switch` statements make the cases a bit hard to follow (at first glance can be hard to tell whether a case is associated with inner or outer switch) -- can we decomp a bit? Would also help readability since this function is pretty large/dense

distgo/docker/docker_push.go
```diff
@@ -78,6 +87,104 @@ func runSingleDockerPush(
...
+		// 1. image index pointing to image indexes -> dereference top-level index and push image index
+		// 2. image index (with multiple platforms) pointing to manifests -> push image index
+		// 3. image index (without multiple platforms) pointing to manifests -> read image from index and push manifest
+		// 4. manifest -> re-extract image from tarball and push manifest :(
+		idxManifest, err := index.IndexManifest()
+		if err != nil {
+			return errors.Wrap(err, "failed to read index manifest")
+		}
+		switch idxManifest.MediaType {
+		case types.OCIImageIndex:
```
nmiyake, line 119: Don't see a `default` case -- is an unknown media type fine to be a no-op, or is it a case that should never happen? If it's the latter, should add default handling that throws an error.

distgo/docker/docker_push.go
```diff
@@ -78,6 +87,104 @@ func runSingleDockerPush(
...
+			// check the type of the inner manifests
+			switch idxManifest.Manifests[0].MediaType {
+			case types.OCIImageIndex:
```
nmiyake, line 122: Same question as previous switch statement: if `MediaType` doesn't match any case, is that an acceptable case where a noop is the correct result, or should there be a `default` case that throws on unrecognized media type?

nmiyake (changes requested):

distgo/docker/docker_push.go
andybradshaw: Added `defaults`, should definitely be errors

distgo/docker/docker_push.go
andybradshaw: Added

distgo/docker/docker_push.go
andybradshaw: Added a `manifestMetadata` type, which should do some validation of index manifest (e.g. has non-zero manifests, all inner manifest media types match) and extracts the "has platform information" check.

It's still a bit dense, might be worth pulling the `remote.Write` and `remote.WriteIndex` into an interface so we can mock it/test this?

changelog/@unreleased/pr-339.v2.yml
andybradshaw: Updated

distgo/docker/docker_push.go
```diff
@@ -78,6 +88,159 @@ func runSingleDockerPush(
...
+	case types.OCIManifestSchema1:
+		if manifestMetadata.containsPlatformInfo {
+			// if we have platform information, we should push our current image index
+			if err := writeIndex(index, ref, productID, dockerID, dryRun, stdout); err != nil {
+				return errors.Wrapf(err, "failed to write image index for tag %s of configuration %s for product %s", ref, dockerID, productID)
+			}
+			return nil
+		}
+
+		if len(idxManifest.Manifests) != 1 {
```
nmiyake, line 158: Line 141 also accesses `idxManifest.Manifests[0]`.

If we always expect to exist, would move this check up to line 137 and then have `manifest := idxManifest.Manifests[0]` and use that variable in the rest of this function

nmiyake: Ah wait it looks like this check is already done in `manifestMetadataFromIndexManifest`, so it's not required here? Or that only checks `>0` so we should keep this semantically as mentioned above?

nmiyake (changes requested):

distgo/docker/docker_push.go
andybradshaw: Yeah, this is a slightly different semantic check. I'm not sure how we'd ever produce an image index with multiple manifests without including platform information... but if we run into a scenario where that happened, I'd rather bail out than publish any of the contained manifests.

nmiyake: OK yes, makes sense in that this is also checking if there are more than 1. To clarify, in the `types.OCIImageIndex`, could the size be different than 1 and it be OK?

Trying to understand why this check is performed here, since it doesn't seem like the logic between lines 154 - 160 would modify this value (would expect this check either at the beginning of the `case` block if it's just for this case, or before the entire `switch` block if expected for all cases)

distgo/docker/docker_push.go
```diff
@@ -78,6 +88,163 @@ func runSingleDockerPush(
...
+	if len(mediaTypes) != 1 {
+		return manifestMetadata{}, errors.Errorf("image index contained mixed inner manifest types: %s", maps.Keys(mediaTypes))
+	}
+
+	mediaType := maps.Keys(mediaTypes)[0]
+	switch mediaType {
+	case types.OCIImageIndex:
+		return manifestMetadata{mediaType: mediaType, containsPlatformInfo: false}, nil
+	case types.OCIManifestSchema1:
+		platformInfo := indexManifest.Manifests[0].Platform != nil
```
nmiyake, line 234: Can you rename this variable `hasPlatformInfo` to make it clearer that it's a boolean value?

Based on the logic in this function, it seems like it's okay if there are multiple manifests as long as they are all of the same `MediaType`. In that scenario, why is it OK to check `Platform != nil` just for the 1st element? Is this value set/not being set guaranteed to be the same for all mainfests?

distgo/docker/docker_push.go
```diff
@@ -78,6 +88,163 @@ func runSingleDockerPush(
...
+	// check the type of the inner manifests
+	manifestMetadata, err := manifestMetadataFromIndexManifest(idxManifest)
+	if err != nil {
+		return errors.Wrap(err, "encountered unexpected index manifest state")
+	}
+
+	switch manifestMetadata.mediaType {
+	case types.OCIImageIndex:
+		// if we have an image index, go one level down and push that
+		innerIndex, err := index.ImageIndex(idxManifest.Manifests[0].Digest)
```
nmiyake, line 145: `idxManifest.Manifests[0]` is referenced here and 3 other places (line 147, 165, 167). Is it not possible to extract this once at line 141 and comment as to what it is/why it's safe (that function before checks there is at least/only 1 element)? Would make it easier to be sure that the access is safe.

nmiyake (approved):
Still have a few questions/feedback, but all of those items are purely stylistic, so providing approval to unblock.

distgo/docker/docker_push.go
andybradshaw: Pulled this into the manifestMetadata struct, so we only read from the index manifest's manifest list in one place.

distgo/docker/docker_push.go
andybradshaw: Renamed and updated to perform this check for all the manifests we iterate over. Was previously making the assumption that if the media type was the same, we'd write the same metadata fields for each of the manifests (platforms in this case). While that's true in practice for how docker is building things now, it's not guaranteed in the spec, so probably safer to confirm that we've got platform information for all manifests.

distgo/docker/docker_push.go
andybradshaw: Yes, when we have both image indexes and image manifests with platform information, we're fine with 1+ manifests. We could invert the platform info check and combine it with this if that makes it more explicit that we only care about there being exactly one when we don't have platform info.

nmiyake (approved):
Thanks, I think the code is much more readable after the refactors and this looks good!

PR #453, opened 2024-02-28
Improvement: Add support for overriding product name in config

asanderson15:
==COMMIT_MSG==
Add support for overriding product name in config
==COMMIT_MSG==

- This enables individual disters and publishers to respect this configuration override in follow-on PRs

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/453)

### DIFF
--- a/distgo/config/configproduct.go
+++ b/distgo/config/configproduct.go
@@ -102,7 +102,7 @@ func (cfg *ProductConfig) ToParam(productID distgo.ProductID, scriptIncludes str
 	}
 	return distgo.ProductParam{
 		ID:                     productID,
-		Name:                   cfg.ProductName,
+		Name:                   getConfigStringValue(cfg.Name, defaultCfg.Name, string(productID)),
 		Build:                  buildParam,
 		Run:                    runParam,
 		Dist:                   distParam,
--- a/distgo/config/configproduct.go
+++ b/distgo/config/configproduct.go
@@ -102,6 +102,7 @@ func (cfg *ProductConfig) ToParam(productID distgo.ProductID, scriptIncludes str
 	}
 	return distgo.ProductParam{
 		ID:                     productID,
+		Name:                   cfg.ProductName,
 		Build:                  buildParam,
 		Run:                    runParam,
 		Dist:                   distParam,
--- a/distgo/config/internal/v0/configproduct.go
+++ b/distgo/config/internal/v0/configproduct.go
@@ -20,6 +20,10 @@ import (
 
 // ProductConfig represents user-specified configuration on how to build a specific product.
 type ProductConfig struct {
+	// ProductName specifies the product name for the product. If not specified,
+	// the product ID will generally be used.
+	ProductName *string `yaml:"product-name,omitempty"`
+
 	// Build specifies the build configuration for the product.
 	Build *BuildConfig `yaml:"build,omitempty"`
 
--- a/distgo/config/internal/v0/configproduct.go
+++ b/distgo/config/internal/v0/configproduct.go
@@ -20,9 +20,11 @@ import (
 
 // ProductConfig represents user-specified configuration on how to build a specific product.
 type ProductConfig struct {
-	// ProductName specifies the product name for the product. If not specified,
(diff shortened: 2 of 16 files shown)


distgo/config/internal/v0/configproduct.go
```diff
@@ -20,6 +20,10 @@ import (
 
 // ProductConfig represents user-specified configuration on how to build a specific product.
 type ProductConfig struct {
+	// ProductName specifies the product name for the product. If not specified,
+	// the product ID will generally be used.
+	ProductName *string `yaml:"product-name,omitempty"`
+
```
nmiyake, line 26: For consistency, I would like this to resemble https://github.com/palantir/distgo/blob/7827e2e713023fb19b1b12fa98c6244036acd117/distgo/config/internal/v0/configbuild.go#L22-L28 more closely (although it doesn't need to be an actual template):

```suggestion
	// Name is the name of the product. This value should be used by tasks that need to use
	// a product name in their rendered configuration or output.
	//
	// If a value is not specified, the value of ProductID is used as the default value.
	Name *string `yaml:"name,omitempty"`
```

distgo/config/configproduct.go
```diff
@@ -102,6 +102,7 @@ func (cfg *ProductConfig) ToParam(productID distgo.ProductID, scriptIncludes str
 	}
 	return distgo.ProductParam{
 		ID:                     productID,
+		Name:                   cfg.ProductName,
```
nmiyake, line 105: I think it will be much cleaner/clearer if `Name` defaults to `ID` if unspecified. 

```suggestion
		Name:                   getConfigStringValue(cfg.Name, defaultCfg.Name, productID),
```

distgo/paramproduct.go
```diff
@@ -23,6 +23,9 @@ type ProductParam struct {
 	// the configuration.
 	ID ProductID
 
+	// Name is the name of this product. If not specified, the ID is generally used.
+	Name *string
```
nmiyake, line 27: Prefer making this non-optional and explicit:

```suggestion
	// Name is the name of this product. If it is not specified in configuration, defaults to the
	// same value as ID. The Name field should be used for operations that require the product's
	// name for configuration or rendering. The field is intended to be used in cases when the
	// product name used for these purposes cannot be the ID -- for example, if there are different
	// products that want to render output with the same logical name.
	Name string
```

distgo/paramproduct.go
```diff
@@ -68,6 +71,7 @@ func (p *ProductParam) AllDependenciesSortedIDs() []ProductID {
 
 type ProductOutputInfo struct {
 	ID                ProductID          `json:"productId"`
+	Name              *string            `json:"productName,omitempty"`
```
nmiyake, line 74: Remove `*` and `omitempty` per above and for consistency.

distgo/paramproduct.go
```diff
@@ -106,9 +110,17 @@ func (p *ProductParam) ToProductOutputInfo(version string) (ProductOutputInfo, e
...
+		Name:              p.Name,
 		BuildOutputInfo:   buildOutputInfo,
 		DistOutputInfos:   distOutputInfos,
 		PublishOutputInfo: publishOutputInfo,
 		DockerOutputInfos: dockerOutputInfos,
 	}, nil
 }
+
+func (p ProductOutputInfo) ProductName() string {
+	if p.Name != nil {
```
nmiyake, line 122: After change above, can just return `p.Name`

nmiyake (changes requested):
Requests for changes for consistency.

I think it would also be most consistent/correct to update [`ProductTemplateFunction`](https://github.com/palantir/distgo/blob/7827e2e713023fb19b1b12fa98c6244036acd117/distgo/template.go#L27) to use `Name` rather than `ProductID` (and possibly to add/expose a `ProductIDTemplateFunction` that specifically renders the ID).

I believe this would be most conceptually correct, while also preserving current behavior as long as default is to set `ProductName` to `ProductID` value.

nmiyake (approved):
Thanks, this looks good to me!

PR #454, opened 2024-02-29
Improvement: Handle legacy plugins that do not include product name

asanderson15:
==COMMIT_MSG==
Handle legacy plugins that do not include product name
==COMMIT_MSG==

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/454)

### DIFF
--- a/distgo/paramproduct.go
+++ b/distgo/paramproduct.go
@@ -15,6 +15,7 @@
 package distgo
 
 import (
+	"encoding/json"
 	"sort"
 )
 
@@ -80,6 +81,17 @@ type ProductOutputInfo struct {
 	DistOutputInfos   *DistOutputInfos   `json:"distOutputInfos"`
 	PublishOutputInfo *PublishOutputInfo `json:"publishOutputInfo"`
 	DockerOutputInfos *DockerOutputInfos `json:"dockerOutputInfos"`
+}
+
+func (p *ProductOutputInfo) UnmarshalJSON(bytes []byte) error {
+	if err := json.Unmarshal(bytes, p); err == nil {
+		return err
+	}
+	// Handle legacy distgo plugins that do not include the Name field yet
+	if p.Name == "" {
+		p.Name = string(p.ID)
+	}
+	return nil
 }
 
 func (p *ProductParam) ToProductOutputInfo(version string) (ProductOutputInfo, error) {
--- a/distgo/paramproduct.go
+++ b/distgo/paramproduct.go
@@ -84,13 +84,15 @@ type ProductOutputInfo struct {
 }
 
 func (p *ProductOutputInfo) UnmarshalJSON(bytes []byte) error {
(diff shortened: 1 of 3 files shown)


distgo/paramproduct.go
```diff
@@ -82,6 +83,17 @@ type ProductOutputInfo struct {
 	DockerOutputInfos *DockerOutputInfos `json:"dockerOutputInfos"`
 }
 
+func (p *ProductOutputInfo) UnmarshalJSON(bytes []byte) error {
+	if err := json.Unmarshal(bytes, p); err == nil {
```
nmiyake, line 87: If you take this approach, pretty sure you have to use a type alias or something -- otherwise, this would recurse infinitely

nmiyake: Concretely, I think something like this would work:

```
func (p *ProductOutputInfo) UnmarshalJSON(data []byte) error {
	type productOutputInfoAlias ProductOutputInfo
	var target productOutputInfoAlias
	if err := json.Unmarshal(data, &target); err != nil {
		return err
	}
	if target.Name == "" {
		target.Name = string(target.ID)
	}
	*p = ProductOutputInfo(target)
	return nil
}
```

Also, this should be cleaned up a bit (assert instead of println etc.), but we should have a test to verify behavior:

```
func Test_ProductOutputInfo_MarshalJSON(t *testing.T) {
	bytes, err := json.Marshal(ProductOutputInfo{
		ID:   "test-id",
		Name: "test-name",
	})
	require.NoError(t, err)
	fmt.Println(string(bytes))

	var got ProductOutputInfo
	err = json.Unmarshal(bytes, &got)
	require.NoError(t, err)
	fmt.Printf("%+v\n", got)

	bytes, err = json.Marshal(ProductOutputInfo{
		ID: "test-id",
	})
	require.NoError(t, err)
	fmt.Println(string(bytes))

	var got2 ProductOutputInfo
	err = json.Unmarshal(bytes, &got2)
	require.NoError(t, err)
	fmt.Printf("%+v\n", got2)
}
```

nmiyake (changes requested):

distgo/paramproduct_test.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"encoding/json"
+	"testing"
+
+	"github.com/palantir/distgo/distgo"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestUnmarshalProductTaskOutputInfo(t *testing.T) {
```
nmiyake, line 26: Can you put these test cases in a loop or have them in separate tests? Want to ensure that there's no cross-contamination between variables for the different test runs (in cases where something does go wrong, since `ptoi` is shared between both unmarshals, we could get in strange states)

nmiyake (changes requested):

nmiyake (approved):
Thanks!

PR #483, opened 2024-06-17
fix: Handle go1.21 changes in packages.Load

bmoylan:
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/483)


distgo/build/imports/imports.go
```diff
@@ -17,28 +17,27 @@ package imports
...
 	"strings"
 
 	"github.com/pkg/errors"
 	"golang.org/x/tools/go/packages"
 )
 
-// GoFiles is a map from package paths to the names of the buildable .go source files (.go files excluding Cgo and test
+// GoFiles is a list of absolute paths to all buildable .go source files (.go files excluding Cgo and test
 // files) in the package.
-type GoFiles map[string][]string
```
bmoylan, line 28: These map keys are unused and were difficult to get consistent between module and nonmodule mode, so I changed this to a flat slice of file paths.

PR #586, opened 2025-05-22
Pass the --platform arg with docker output type

sonnysasaka:
## Before this PR
Right now cross platform build result has to be taken from the OCI output.

## After this PR
==COMMIT_MSG==
For convenience, we should allow cross platform build for docker output type as well and for this we need to pass the `--platform` arg.

If multiple platforms are specified, only the native platform will be picked. If one platform is specified, this platform will be picked even for cross platform.
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/586)

### DIFF
--- a/changelog/@unreleased/pr-586.v2.yml
+++ b/changelog/@unreleased/pr-586.v2.yml
@@ -1,0 +1,5 @@
+type: improvement
+improvement:
+  description: Pass the --platform arg with docker output type.
+  links:
+  - https://github.com/palantir/distgo/pull/586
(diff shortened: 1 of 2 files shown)


palantirtech:
Thanks for your interest in [palantir/distgo](https://github.com/palantir/distgo), @sonnysasaka! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nmiyake (approved):

PR #594, opened 2025-06-10
Revert "Pass the --platform arg with docker output type (#586)"

sonnysasaka:
This reverts commit 1b1944f88d53fbd7398ac2e08fc0229b342685d3.

Reason for revert:
This breaks systems that don't use containerd image store (https://github.com/docker/buildx/issues/59).

## After this PR

==COMMIT_MSG==
==COMMIT_MSG==


changelog/@unreleased/pr-594.v2.yml
```diff
@@ -0,0 +1,5 @@
+type: fix
+improvement:
```
nmiyake, line 2: Needs to be `fix:` to make this a valid union type

PR #605, opened 2025-08-06
Support --insecure flag for docker push

sonnysasaka:
## Before this PR
`./godelw docker push --repository localhost:5000` does not work

## After this PR
Support `--insecure` flag that conveniently allows docker push to development local registry which is usually plain HTTP.

==COMMIT_MSG==
Support --insecure flag for docker push
==COMMIT_MSG==


distgo/docker/docker_push.go
```diff
@@ -86,23 +87,30 @@ func runSingleDockerPush(
...
 
 	for _, tag := range productTaskOutputInfo.Product.DockerOutputInfos.DockerBuilderOutputInfos[dockerID].RenderedTags {
-		ref, err := name.ParseReference(tag)
+		var ref name.Reference
+		var err error
+		if insecure {
+			ref, err = name.ParseReference(tag, name.Insecure)
+		} else {
+			ref, err = name.ParseReference(tag)
+		}
```
nmiyake, line 113: I think the following construction is simpler/more straightforward:

```suggestion
		var opts []name.Option
		if insecure {
			opts = append(opts, name.Insecure)
		}
		ref, err := name.ParseReference(tag, opts...)
```

distgo/docker/docker_push.go
```diff
@@ -118,11 +126,11 @@ func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTas
 		}
 		switch idxManifest.MediaType {
 		case types.OCIImageIndex:
-			if err := handleImageIndex(index, idxManifest, ref, productID, dockerID, dryRun, stdout); err != nil {
+			if err := handleImageIndex(index, idxManifest, ref, productID, dockerID, dryRun, insecure, stdout); err != nil {
```
nmiyake, line 129: I believe that once the `insecure` parameter is read and applied to `ref` in the block above, it is no longer read/used in any of these functions, so you should just remove `insecure` as a parameter and not pass it in where it is not used

nmiyake (changes requested):
Thanks -- ok with feature, but feedback on specifics

nmiyake (approved):

PR #792, opened 2026-02-24
feature: Integrate govulncheck to produce VEX documents

gcampbell12:
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/792)


nmiyake:
Thanks for the PR -- this prototype was very helpful to see the approach and explore the problem space. After consideration, we are doing most of this work, but decided to do it as part of the CI job rather than in the `distgo` plugin, so closing this PR.

PR #875, opened 2026-06-11
fix: invoke bash instead of execing build scripts directly

OpenSourceContributorName:
## Before this PR
When build scripts are run, they are execed directly after they are written which leaves a window of time where a child process inherits the FDs of the parent and blocks another child from spawning leading to an EXTBUSY error. 

See https://github.com/golang/go/issues/22315 for more discussion on this kind of issue.

## After this PR
Invoke Bash and allow Bash to run the script removing the race around writing and execing the same file.

==COMMIT_MSG==
fix: invoke bash instead of execing build scripts directly
==COMMIT_MSG==

## Possible downsides?
None.

### DIFF
--- a/distgo/script.go
+++ b/distgo/script.go
@@ -80,7 +80,7 @@ func WriteAndExecuteScript(projectInfo ProjectInfo, script string, additionalEnv
 			env = append(env, fmt.Sprintf("%v=%v", k, v))
 		}
 
-		cmd := exec.Command(tmpFile)
+		cmd := exec.Command("bash", tmpFile)
 		cmd.Dir = projectInfo.ProjectDir
 		cmd.Env = env
 		cmd.Stdout = stdOut
--- a/distgo/script.go
+++ b/distgo/script.go
@@ -25,6 +25,8 @@ import (
 	"github.com/pkg/errors"
 )
 
+const bashBinaryPath = "/bin/bash"
+
 func WriteScript(projectInfo ProjectInfo, script string) (name string, cleanup func() error, rErr error) {
 	tmpFile, err := os.CreateTemp(projectInfo.ProjectDir, "")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


palantirtech:
Thanks for your interest in [palantir/distgo](https://github.com/palantir/distgo), @OpenSourceContributorName! Before we can accept your pull request, you need to sign our contributor license agreement - just visit https://cla.palantir.com/ and follow the instructions. Once you sign, I'll automatically update this pull request.

nmiyake (changes requested):
Thanks for tracking down the related thread -- I had noticed this intermittently, but didn't know about the underlying issue!

Although this will likely work in most cases, I think it would be preferable to solve/work around this in pure Go rather than invoking `bash`.

For now, let's use an approach like https://go-review.googlesource.com/c/go/+/790840 (or the others referenced on the thread) to just do a time-based retry to attempt to work around the issue. We should also add a comment to reference the thread/issue for more context.

OpenSourceContributorName:
@nmiyake Will defer to you but important context here is that the scripts seem to be 100% bash when we [write them](https://github.com/palantir/distgo/blob/14ddeeddb727f1a86852952d7cd001c97ad574a4/distgo/script.go#L46) so we would be invoking bash either way.

nmiyake:
Thanks, that's a great point. In sync on the approach, but would just like a few modifications to be extra safe:

Let's factor out `/bin/bash` into a constant (maybe this should be `/usr/bin/env bash`, but we don't have to do that for now) and then use that both in the script writing logic and in this invocation logic. That way we can ensure that the script running and our running will always agree on the `bash` to use (and if we later modify the shell or something, we can ensure that both locations are updated).

Also, let's add a comment to the new exec line that references https://github.com/golang/go/issues/22315 and states that this is done to work around the "text file busy" error.

nmiyake (approved):

PR #881, opened 2026-06-23
Add the product-maven-coord task

nmiyake:
## After this PR

==COMMIT_MSG==
Adds a task the prints the maven coordinates for products.
==COMMIT_MSG==

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/881)

### DIFF
--- a/distgo/productmavencoord/productmavencoord.go
+++ b/distgo/productmavencoord/productmavencoord.go
@@ -1,0 +1,57 @@
+// Copyright 2026 Palantir Technologies, Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package productmavencoord
+
+import (
+	"fmt"
+	"io"
+	"maps"
+	"slices"
+
+	"github.com/palantir/distgo/distgo"
+	"github.com/palantir/distgo/publisher"
+)
+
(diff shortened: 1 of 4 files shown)


distgo/productmavencoord/productmavencoord.go
```diff
@@ -0,0 +1,57 @@
...
+		productIDs = slices.Sorted(maps.Keys(projectParam.Products))
+	} else {
+		// otherwise, filter products to only those specified
+		for _, productID := range specifiedProductIDs {
+			if _, ok := projectParam.Products[productID]; !ok {
+				continue
+			}
+			productIDs = append(productIDs, productID)
+		}
+	}
```
tabboud: This could be simplified using the following, which other tasks use. The only caveat is this will error out on non-existent products, which deviates from the implementation here which just filters them out. Was that an intentional decision?

```go
productParams, err := distgo.ProductParamsForProductArgs(projectParam.Products, specifiedProductIDs...)
if err != nil {
	return err
}
```

nmiyake: Thanks, good catch! I did the filtering out because I thought that matched what we did in other tasks like this (for `publish`), but don't feel strongly about it, so think that using the library function here probably makes more sense.

nmiyake:
Thanks for the review! Addressed feedback.

distgo/productmavencoord/productmavencoord.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func Run(projectInfo distgo.ProjectInfo, projectParam distgo.ProjectParam, specifiedProductIDs []distgo.ProductID, stdout io.Writer) error {
+	productParams, err := distgo.ProductParamsForProductArgs(projectParam.Products, specifiedProductIDs...)
+	if err != nil {
+		return err
+	}
+	for _, productParam := range productParams {
+		productID := productParam.ID
+		productTaskOutputInfo, err := distgo.ToProductTaskOutputInfo(projectInfo, projectParam.Products[productID])
```
tabboud: You can just use the `productParam` loop var instead of indexing

```suggestion
		productTaskOutputInfo, err := distgo.ToProductTaskOutputInfo(projectInfo, productParam)
```

tabboud (commented):
Now that we error out on non-existent productIDs, you will need to update the tests accordingly.

nmiyake:
Thanks! Addressed feedback and fixed tests.

tabboud (approved):

PR #900, opened 2026-07-15
Fix: Github publisher creates releases as draft until assets are uploaded

tabboud:
## Before this PR

GitHub's immutable-releases feature ([docs](https://docs.github.com/en/code-security/concepts/supply-chain-security/immutable-releases)) does not allow modifying or uploading assets after the release has been created. The `github` publisher creates a release and then uploads dist artifacts to it afterward, but now that a release can be immutable the after it's published, every upload will fail with a `422 Cannot upload assets to an immutable release`. See this GH action run from distgo https://github.com/palantir/distgo/actions/runs/29429428387/job/87400271522
The release/tag ends up existing with zero assets, breaking excavators and prevents updates for the tooling.

The last known good release was [v1.96.0](https://github.com/palantir/distgo/releases/tag/v1.96.0) (on 2/26/26) and has been broken since (we're now on [v1.103.0](https://github.com/palantir/distgo/releases/tag/v1.103.0)). This is likely impacting ALL godel plugins/assets as well.

## After this PR

==COMMIT_MSG==
GitHub publisher creates release as draft until assets are uploaded

The GitHub publisher now creates the release as a draft, uploads all dist artifacts to it, and only publishes (un-drafts) the release once every upload has succeeded. This allows the publisher to work when GitHub's immutable-releases feature is enabled.
==COMMIT_MSG==

## Possible downsides?

If the upload step fails partway through, the release is now left as a draft instead of a partially published release. This is a behavior change, but strictly safer since a draft can be retried/completed, whereas a broken published release has to be fixed manually.

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/900)

### DIFF
--- a/publisher/github/publisher.go
+++ b/publisher/github/publisher.go
@@ -155,8 +155,11 @@ func (p *githubPublisher) RunPublish(productTaskOutputInfo distgo.ProductTaskOut
 
 	var releaseRes *github.RepositoryRelease
 	if !dryRun {
+		// create the release as a draft since GitHub's immutable-releases feature rejects asset uploads to a
+		// non-draft release, so uploads must happen before the release is published.
 		releaseRes, _, err = client.Repositories.CreateRelease(context.Background(), cfg.Owner, cfg.Repository, &github.RepositoryRelease{
 			TagName: new(releaseVersion),
+			Draft:   new(true),
 		})
 		if err != nil {
 			// newline to complement "..." output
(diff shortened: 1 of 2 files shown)


publisher/github/publisher.go
```diff
@@ -186,6 +189,19 @@ func (p *githubPublisher) RunPublish(productTaskOutputInfo distgo.ProductTaskOut
...
+	if !dryRun && releaseRes.GetDraft() {
+		distgo.PrintOrDryRunPrint(stdout, fmt.Sprintf("Publishing GitHub release %s for %s/%s...", releaseVersion, cfg.Owner, cfg.Repository), dryRun)
+		if _, _, err := client.Repositories.EditRelease(context.Background(), cfg.Owner, cfg.Repository, releaseRes.GetID(), &github.RepositoryRelease{
+			Draft: new(false),
+		}); err != nil {
+			_, _ = fmt.Fprintln(stdout)
+			return errors.Wrapf(err, "failed to publish GitHub release %s for %s/%s after uploading assets", releaseVersion, cfg.Owner, cfg.Repository)
+		}
+		_, _ = fmt.Fprintln(stdout, "done")
+	}
```
nmiyake, line 206: I think it would be slightly better to print operation as part of dry run

```suggestion
	// publish the release now that all assets have been uploaded. Only necessary if the release was created as a
	// draft above (a pre-existing, already-published release found via the "already_exists" path above does not need this step).
	if releaseRes.GetDraft() {
		distgo.PrintOrDryRunPrint(stdout, fmt.Sprintf("Publishing GitHub release %s for %s/%s...", releaseVersion, cfg.Owner, cfg.Repository), dryRun)
		if !dryRun {
				if _, _, err := client.Repositories.EditRelease(context.Background(), cfg.Owner, cfg.Repository, releaseRes.GetID(), &github.RepositoryRelease{
			Draft: new(false),
				}); err != nil {
					_, _ = fmt.Fprintln(stdout)
					return errors.Wrapf(err, "failed to publish GitHub release %s for %s/%s after uploading assets", releaseVersion, cfg.Owner, cfg.Repository)
				}
		}
		_, _ = fmt.Fprintln(stdout, "done")
	}
```

nmiyake (changes requested):
Minor request for printing in dry-run mode

nmiyake (approved):

PR #904, opened 2026-07-17
Emit the cross-product FROM build-context at the Docker task level

bmoylan:
A cross-product `FROM <dep tag>` resolves from the dependency's on-disk OCI layout with no registry, using a small buildx "wrapper" layout written next to the image. That wrapper was written inside the default builder's extractToOCILayout (#895), which had two gaps:

  - Re-layering builders (e.g. the chunkah asset) rewrite the OCI layout after building, clobbering a wrapper the builder had written, so a chunked image could never be a local FROM base.
  - The built-in `default` builder is daemon-only and writes no OCI layout, so it produced no wrapper at all.

Move wrapper emission into the Docker build task, run after the builder returns against whatever OCI layout it left on disk. It is now builder-agnostic: any builder that leaves an OCI layout can serve as a local FROM base, including re-layering builders. The producer is now distgo.WriteDockerBuildContextLayout (self-contained -- it derives the descriptor from the layout's index.json rather than receiving it from the builder); buildcontext.go keeps only the consumer side.

Also make the built-in `default` builder emit an OCI layout in addition to loading the host-arch image into the daemon, so a plain `type: default` product can be a local base without opting into a heavier builder (`godel docker push` already prefers the OCI layout when present).

Incidental cleanup in RunDockerBuild: name the shared args `baseArgs` and give each output its own slices.Clone, replacing append-aliasing that only worked because exec.Command copies its args.

Update to feature added in #895

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/904)

### DIFF
--- a/distgo/docker/docker_build.go
+++ b/distgo/docker/docker_build.go
@@ -239,7 +239,26 @@ func runSingleDockerBuild(
 
 	distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("Running Docker build for configuration %s of product %s...", dockerID, productID), dryRun)
 	// run the Docker build task
-	return dockerBuilderParam.DockerBuilder.RunDockerBuild(dockerID, productTaskOutputInfo, verbose, dryRun, stdout)
+	if err := dockerBuilderParam.DockerBuilder.RunDockerBuild(dockerID, productTaskOutputInfo, verbose, dryRun, stdout); err != nil {
+		return err
+	}
+
+	// If the builder produced an OCI layout, write the buildx build-context wrapper so a dependent product's
+	// "FROM <this image's tag>" can resolve from the on-disk layout with no registry. This is done here at the task
+	// level -- rather than inside the builder -- so it works for every builder that leaves an OCI layout, including
+	// re-layering builders (e.g. the chunkah asset) that rewrite the layout after building and would otherwise clobber
+	// a wrapper the builder wrote itself. Builders that produce no OCI layout (daemon-only) leave no index.json and are
+	// skipped: they cannot serve as a local FROM base.
+	if !dryRun {
+		ociDir := productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)
+		if _, err := os.Stat(filepath.Join(ociDir, "index.json")); err == nil {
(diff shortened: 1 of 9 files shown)


dockerbuilder/defaultdockerbuilder/dockerbuilder.go
```diff
@@ -58,7 +59,9 @@ func NewDefaultDockerBuilder(buildArgs []string, buildArgsScript string) distgo.
 	return &DefaultDockerBuilder{
 		BuildArgs:       buildArgs,
 		BuildArgsScript: buildArgsScript,
-		OutputType:      DockerDaemon,
+		// Produce an OCI layout (the reproducible published artifact, and the on-disk base a dependent product's FROM
+		// resolves against) and load the host-arch image into the local daemon for `docker run`.
+		OutputType: OCILayout | DockerDaemon,
```
tabboud, line 64: Now that the default docker builder encodes `OCILayout`, the check on line 114 will always be true, however this relies on the fact that there is a corresponding dist artifact which I don't believe is always the case.

distgo/docker/docker_build.go
```diff
@@ -239,7 +239,26 @@ func runSingleDockerBuild(
...
+	}
+
+	// If the builder produced an OCI layout, write the buildx build-context wrapper so a dependent product's
+	// "FROM <this image's tag>" can resolve from the on-disk layout with no registry. This is done here at the task
+	// level -- rather than inside the builder -- so it works for every builder that leaves an OCI layout, including
+	// re-layering builders (e.g. the chunkah asset) that rewrite the layout after building and would otherwise clobber
+	// a wrapper the builder wrote itself. Builders that produce no OCI layout (daemon-only) leave no index.json and are
+	// skipped: they cannot serve as a local FROM base.
+	if !dryRun {
+		ociDir := productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)
```
tabboud: Similar to the other comment, this could technically produce `""` if there was no dist

dockerbuilder/defaultdockerbuilder/dockerbuilder.go
bmoylan: Thanks, fixed to skip when empty

tabboud (approved):

PR #915, opened 2026-07-22
Fix: Github publisher finalizes a shared release after all asset uploads

tabboud:
## Before this PR
Products publishing more than one dist to the same tag fail when GitHub immutable releases are enabled. This was seen on `golangci-lint-palantir` (https://github.com/palantir/golangci-lint-palantir/actions/runs/29871701351/job/88773078087) since the initial product was finalized (making the release immutable) and the second product failed to finalize, resulting in a dangling draft with the second dist artifacts.

## After this PR

==COMMIT_MSG==
Github publisher finalizes a shared release after all asset uploads
==COMMIT_MSG==

`RunPublish` now only creates/reuses a draft and uploads assets. A new `FinalizingPublisher` interface allows publishers to implement the finalizing logic after all assets for every product have been uploaded. In the GitHub case, `FinalizePublish` is used to un-draft the release after all products upload and in the case where 2 products publish under the same tag, the second `FinalizePublish` call will be a noop since a draft release will not be found.

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/915)

### DIFF
--- a/distgo/publish/publish.go
+++ b/distgo/publish/publish.go
@@ -35,21 +35,65 @@ func Products(projectInfo distgo.ProjectInfo, projectParam distgo.ProjectParam, 
 	if err != nil {
 		return err
 	}
+	var publishedProducts []publishedProduct
 	for _, currProduct := range productParams {
-		if err := Run(projectInfo, currProduct, publisher, flagVals, dryRun, stdout); err != nil {
+		publishedProduct, err := publishProduct(projectInfo, currProduct, publisher, flagVals, dryRun, stdout)
+		if err != nil {
 			return err
+		}
+		if publishedProduct != nil {
+			publishedProducts = append(publishedProducts, *publishedProduct)
+		}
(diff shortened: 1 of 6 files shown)


nmiyake:
> the second FinalizePublish call will be a noop since a draft release will not be found.

What does this part mean? Since the finalizer is per-product, if there are multiple products that publish under the same tag, would this then not work?

nmiyake:
Ah OK now I understand -- so basically while all the products are publishing, it's still a draft so all the products can upload/add assets. Then, in the finalize loop, the first product for a tag/release will finalize it (and the release is already in the final/ready state), and although other products that published to the same release will also call "Finalize", it will be a noop because the release has already been finalized, and that's okay because it's still in the same state.

tabboud:
Replaced by https://github.com/palantir/distgo/pull/927

PR #918, opened 2026-07-23
Feature: Batch publish across products sharing a GH release

tabboud:
## Before this PR
`distgo` publishes each product individually through `publisher.RunPublish` with no way for a publisher to coordinate work across products. When multiple products publish to the same GH release, the first products publish call will un-draft the release and when the immutable-releases feature is enabled, the second products publish call will fail since the release cannot be modified after finalizing. This was seen on `golangci-lint-palantir` (https://github.com/palantir/golangci-lint-palantir/actions/runs/29871701351/job/88773078087) since the initial product was finalized (making the release immutable) and the second product failed to finalize, resulting in a dangling draft with the second dist artifacts.

## After this PR

==COMMIT_MSG==
Batch publish across products sharing a GH release
==COMMIT_MSG==

A new optional interface `distgo.BatchPublisher` is introduced which allows the publishers to handle the orchestration of publishing all products. This is what allows the GH publisher to ensure all assets for a release are uploaded before finalizing the release. Retrying a publish now skips assets that are already present on a release in the case of a partial failure. Publishers that do not implement the `BatchPublisher` are unaffected.

This PR alone does not extend support for batch publishing to the asset publishers. This will come in a FLUP.

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/918)

### DIFF
--- a/publisher/github/publisher.go
+++ b/publisher/github/publisher.go
@@ -97,30 +97,13 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
 }
 
 func (p *githubPublisher) RunPublish(productTaskOutputInfo distgo.ProductTaskOutputInfo, cfgYML []byte, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
-	var cfg config.GitHub
-	if err := yaml.Unmarshal(cfgYML, &cfg); err != nil {
-		return errors.Wrapf(err, "failed to unmarshal configuration")
-	}
-	if err := publisher.SetRequiredStringConfigValues(flagVals,
-		githubPublisherAPIURLFlag, &cfg.APIURL,
-		githubPublisherUserFlag, &cfg.User,
-		githubPublisherTokenFlag, &cfg.Token,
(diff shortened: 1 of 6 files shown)


publisher/github/publisher.go
```diff
@@ -252,7 +353,20 @@ func (p *githubPublisher) uploadFileAtPath(client *github.Client, release *githu
...
 	}
 
-	uploadURI, err := uploadURIForProduct(release.GetUploadURL(), path.Base(filePath))
+	assetName := path.Base(filePath)
+	if existingAsset := findReleaseAsset(release, assetName); existingAsset != nil {
+		stat, err := f.Stat()
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to stat artifact %s", filePath)
+		}
+		if int64(existingAsset.GetSize()) != stat.Size() {
```
tabboud, line 362: This is all to try and ensure a re-run of publishing doesn't try to upload an asset that was already uploaded. The `go-github` module we use is quite outdated and does not support checksums, so I opted for name + size verification instead.

tabboud:
Replaced by https://github.com/palantir/distgo/pull/927

PR #919, opened 2026-07-23
Feature: Extend batch publishing to asset publishers

tabboud:
## Before this PR
#918 adds support for batch publishing but only for internal publishers using the Go API. Asset publishers always went through the per-product `RunPublish` flow.

## After this PR

==COMMIT_MSG==
Extend batch publishing to asset publishers
==COMMIT_MSG==

The `publisher.AssetRootCmd` now registers a `run-publish-batch` command on every asset and calls `RunPublishBatch` if the wrapped publisher implements the `distgo.BatchPublisher` interface, otherwise it falls back to looping over all inputs and running `RunPublish` for each one. distgo probes each asset at startup and calls `run-publish-batch --help` to determine if it's supported in order to provide backcompat for older assets.

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/919)


tabboud:
Replaced by #923

PR #923, opened 2026-07-24
Feature: Make publisher.RunPublish batch publish products by default

tabboud:
## Before this PR
`publisher.RunPublish` publishes a single product at a time. #918 and #919 introduced an additive publisher type to support coordinating work across product publishing, however this made the API harder to discover and added extra overhead for users deciding whether they wanted to implement batch publishing or not.

## After this PR

==COMMIT_MSG==
Make publisher.RunPublish batch publish products by default
==COMMIT_MSG==

The `publisher.RunPublish` method now takes in a set of products `[]PublishInput` and the publisher implementations can decide how to orchestrate the work of publishing. Right now, all internal implementations just loop over the inputs and run the single publish command. However, a FLUP PR will adjust the GitHub publisher to coordinate publishing across all products in a single release.

The asset API has been updated to add a new `run-publish-v2` command which just calls `RunPublish` with the full batch of inputs. The legacy `run-publish` command is kept to support backcompat and wraps the single product in the `[]PublishInput` and still calls the batched `RunPublish` method.

## Possible downsides?
- This is a breaking change that will require some updates and coordination with existing asset publishers, such as `godel-conjure-plugin`

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/923)

### DIFF
--- a/distgo/publish/publish.go
+++ b/distgo/publish/publish.go
@@ -35,21 +35,38 @@ func Products(projectInfo distgo.ProjectInfo, projectParam distgo.ProjectParam, 
 	if err != nil {
 		return err
 	}
+
+	publisherType, err := publisher.TypeName()
+	if err != nil {
+		return errors.Wrapf(err, "failed to determine type of publisher")
+	}
+
+	var inputs []distgo.PublishInput
 	for _, currProduct := range productParams {
-		if err := Run(projectInfo, currProduct, publisher, flagVals, dryRun, stdout); err != nil {
+		input, err := preparePublishInput(projectInfo, currProduct, publisherType, dryRun, stdout)
+		if err != nil {
 			return err
 		}
+		if input == nil {
+			continue
+		}
+		inputs = append(inputs, *input)
+	}
+	if len(inputs) == 0 {
+		return nil
+	}
+	if err := publisher.RunPublish(inputs, flagVals, dryRun, stdout); err != nil {
+		return errors.Wrapf(err, "failed to publish products using %s publisher", publisherType)
 	}
 	return nil
 }
 
-// Run executes the publish action for the specified product. Produces both the dist output directory and the dist
-// artifacts for the product. The outputs for the dependent products for the provided product must already exist in the
-// proper locations.
-func Run(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisher distgo.Publisher, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
+// preparePublishInput computes the [distgo.PublishInput] for the specified product, or nil if it has no dist outputs
+// and should be skipped. The outputs for its dependent products must already exist in their proper locations.
+func preparePublishInput(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisherType string, dryRun bool, stdout io.Writer) (*distgo.PublishInput, error) {
 	if productParam.Dist == nil {
 		distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("%s does not have dist outputs; skipping publish", productParam.ID), dryRun)
-		return nil
+		return nil, nil
 	}
 
 	// verify that dist artifacts to publish exists (dist is skipped in dry-run mode, so the
@@ -57,33 +74,27 @@ func Run(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publi
 	if !dryRun {
 		productOutputInfo, err := productParam.ToProductOutputInfo(projectInfo.Version)
 		if err != nil {
-			return errors.Wrapf(err, "failed to compute output info")
+			return nil, errors.Wrapf(err, "failed to compute output info")
 		}
 		for _, currDistID := range productOutputInfo.DistOutputInfos.DistIDs {
 			for _, currArtifactPath := range distgo.ProductDistArtifactPaths(projectInfo, productOutputInfo)[currDistID] {
 				if _, err := os.Stat(currArtifactPath); os.IsNotExist(err) {
-					return errors.Errorf("distribution artifact for product %s with dist %s does not exist at %s", productParam.ID, currDistID, currArtifactPath)
+					return nil, errors.Errorf("distribution artifact for product %s with dist %s does not exist at %s", productParam.ID, currDistID, currArtifactPath)
 				}
 			}
 		}
 	}
 
-	// run publish
 	productTaskOutputInfo, err := distgo.ToProductTaskOutputInfo(projectInfo, productParam)
 	if err != nil {
-		return err
-	}
-	publisherType, err := publisher.TypeName()
-	if err != nil {
-		return errors.Wrapf(err, "failed to determine type of publisher")
+		return nil, err
 	}
 	var publishCfgBytes []byte
 	if productParam.Publish != nil {
 		publishCfgBytes = productParam.Publish.PublishInfo[distgo.PublisherTypeID(publisherType)].ConfigBytes
 	}
-	if err := publisher.RunPublish(productTaskOutputInfo, publishCfgBytes, flagVals, dryRun, stdout); err != nil {
-		return errors.Wrapf(err, "failed to publish %s using %s publisher", productParam.ID, publisherType)
-	}
-
-	return nil
+	return &distgo.PublishInput{
+		ProductTaskOutputInfo: productTaskOutputInfo,
+		ConfigYML:             publishCfgBytes,
+	}, nil
 }
--- a/distgo/publish/publish.go
+++ b/distgo/publish/publish.go
@@ -41,9 +41,9 @@ func Products(projectInfo distgo.ProjectInfo, projectParam distgo.ProjectParam, 
 		return errors.Wrapf(err, "failed to determine type of publisher")
 	}
 
-	var inputs []distgo.PublishInput
+	var inputs []distgo.ProductPublishInfo
 	for _, currProduct := range productParams {
-		input, err := preparePublishInput(projectInfo, currProduct, publisherType, dryRun, stdout)
+		input, err := getProductPublishInfo(projectInfo, currProduct, publisherType, dryRun, stdout)
 		if err != nil {
 			return err
 		}
@@ -61,9 +61,10 @@ func Products(projectInfo distgo.ProjectInfo, projectParam distgo.ProjectParam, 
 	return nil
 }
 
-// preparePublishInput computes the [distgo.PublishInput] for the specified product, or nil if it has no dist outputs
-// and should be skipped. The outputs for its dependent products must already exist in their proper locations.
-func preparePublishInput(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisherType string, dryRun bool, stdout io.Writer) (*distgo.PublishInput, error) {
+// getProductPublishInfo computes and returns the [distgo.ProductPublishInfo] for the specified productParam, or nil
+// if it has no dist outputs and should be skipped. The outputs for any dependent products of the param must already
+// exist in their proper locations.
+func getProductPublishInfo(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisherType string, dryRun bool, stdout io.Writer) (*distgo.ProductPublishInfo, error) {
 	if productParam.Dist == nil {
 		distgo.PrintlnOrDryRunPrintln(stdout, fmt.Sprintf("%s does not have dist outputs; skipping publish", productParam.ID), dryRun)
 		return nil, nil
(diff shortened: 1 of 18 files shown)


publisher/publisher_asset.go
```diff
@@ -54,23 +54,18 @@ func (p *assetPublisher) Flags() ([]distgo.PublisherFlag, error) {
...
 
-	args := []string{runPublishCmdName}
-	args = append(args, "--"+runPublishCmdProductTaskOutputInfoFlagName, string(productTaskOutputInfoJSON))
-	cfgYMLString := string(cfgYML)
-	if cfgYMLString == "" {
-		cfgYMLString = "{}"
-	}
-	args = append(args, "--"+runPublishCmdConfigYMLFlagName, cfgYMLString)
+	args := []string{runPublishV2CmdName}
+	args = append(args, "--"+runPublishV2CmdInputsFlagName, string(inputsJSON))
```
tabboud: I kept this input as a flag that we pass in the set of inputs, however this list could get large given the number of products that need to be published. In practice it's likely that we publish < 10 products at once, so this should be fine. We could switch this over to passing the set of products as stdin which mitigates the size requirement, but I figured this wasn't needed right now.

distgo/publisher.go
```diff
@@ -28,10 +28,17 @@ type Publisher interface {
 	// Flags returns the flags provided by this Publisher.
 	Flags() ([]PublisherFlag, error)
 
-	// RunPublish runs the publish task. When this function is called, the distribution artifacts for the product should
-	// already exist. If dryRun is true, then the task should print the operations that would occur without actually
-	// executing them.
-	RunPublish(productTaskOutputInfo ProductTaskOutputInfo, cfgYML []byte, flagVals map[PublisherFlagName]any, dryRun bool, stdout io.Writer) error
+	// RunPublish runs the publish task for every product in inputs. When this function is called, the distribution artifacts
```
nmiyake:
```suggestion
	// RunPublish runs the publish task to publish the products specified in the provided ProductPublishInfos. When this function is called, the distribution artifacts
```

publisher/internal/publishfixtures/v2/main.go
```diff
@@ -0,0 +1,64 @@
...
+func (p *v2Publisher) Flags() ([]distgo.PublisherFlag, error) {
+	return nil, nil
+}
+
+func (p *v2Publisher) RunPublish(inputs []distgo.PublishInput, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
+	ids := make([]string, 0, len(inputs))
+	for _, input := range inputs {
+		ids = append(ids, string(input.ProductTaskOutputInfo.Product.ID))
+	}
+	sort.Strings(ids)
```
nmiyake: nit, but since `inputs` is already ordered, I think it makes more sense to preserve the provided ordering rather than sorting on the ID (sorting by the ID would make more sense if the input was a map or some other unordered structure and we needed to impose our own ordering on top of it)

publisher/assetapi.go
```diff
@@ -100,7 +102,7 @@ func newRunPublishCmd(publisher distgo.Publisher) *cobra.Command {
 	)
 	runDistCmd := &cobra.Command{
 		Use:   runPublishCmdName,
-		Short: "Runs the publish action",
+		Short: "Runs the publish action for a single product",
```
nmiyake:
```suggestion
		Short: "Runs the publish action for a single ProductTaskOutputInfo",
```

publisher/assetapi.go
```diff
@@ -126,6 +132,44 @@ func newRunPublishCmd(publisher distgo.Publisher) *cobra.Command {
 	return runDistCmd
 }
 
+const (
+	runPublishV2CmdName           = "run-publish-v2"
+	runPublishV2CmdInputsFlagName = "inputs"
+)
+
+// newRunPublishV2Cmd returns the run-publish-v2 command which publishes a batch of products.
```
nmiyake:
```suggestion
// newRunPublishV2Cmd returns the run-publish-v2 command which publishes the provided  []distgo.ProductPublishInfo
```

publisher/publisher_asset.go
nmiyake: Thanks, good flag.

I think this makes sense and agree this is likely safe, but just so that we can get a better sense of this, could you test this out on a few real products to get a general sense of the representative size for a single `distgo.PublishInput` and the limit for argument sizes so we can get a general sense of at least the order of magnitude of the max we would be looking at here?

For max size I'm seeing different things initially (Google saying 256kb for macOS, but seeing 1048576, which is 1,048kb, but also seeing 128kb max for a single argument), so might also be helpful to do a quick experiment to check that on macos and linux.

Basically agree that if we can support 10-20 products, I think we're in the clear (and if it's 50-100+, then I think we're totally fine). However, if it's more like 6-10, then I think it may be worth going with `stdin` at the outset just to be safe.

distgo/publisher.go
```diff
@@ -28,10 +28,17 @@ type Publisher interface {
...
+	// actually executing them.
+	RunPublish(inputs []PublishInput, flagVals map[PublisherFlagName]any, dryRun bool, stdout io.Writer) error
+}
+
+// PublishInput contains the product output information and product-specific configuration for one product's
+// publish operation.
+type PublishInput struct {
+	ProductTaskOutputInfo ProductTaskOutputInfo
+	ConfigYML             []byte
 }
```
nmiyake: Some suggestions that I think would make the understandability of this a bit clearer.

```suggestion
// ProductPublishInfo contains the product output information and product-specific publisher configuration for one product's
// publish operation for a particular publisher.
type ProductPublishInfo struct {
	ProductTaskOutputInfo ProductTaskOutputInfo
	PublisherConfigYML             []byte
}
```

distgo/publish/publish.go
```diff
@@ -35,55 +35,66 @@ func Products(projectInfo distgo.ProjectInfo, projectParam distgo.ProjectParam,
...
 	return nil
 }
 
-// Run executes the publish action for the specified product. Produces both the dist output directory and the dist
-// artifacts for the product. The outputs for the dependent products for the provided product must already exist in the
-// proper locations.
-func Run(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisher distgo.Publisher, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
+// preparePublishInput computes the [distgo.PublishInput] for the specified product, or nil if it has no dist outputs
+// and should be skipped. The outputs for its dependent products must already exist in their proper locations.
+func preparePublishInput(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisherType string, dryRun bool, stdout io.Writer) (*distgo.PublishInput, error) {
```
nmiyake:
```suggestion
// getProductPublishInfo computes and returns the [distgo.ProductPublishInfo] for the specified productParam, or nil if it has no dist outputs
// and should be skipped. The outputs for any dependent products of the param must already exist in their proper locations.
func getProductPublishInfo(projectInfo distgo.ProjectInfo, productParam distgo.ProductParam, publisherType string, dryRun bool, stdout io.Writer) (*distgo.PublishInput, error) {
```

publisher/assetapi.go
```diff
@@ -126,6 +132,44 @@ func newRunPublishCmd(publisher distgo.Publisher) *cobra.Command {
...
+func newRunPublishV2Cmd(publisher distgo.Publisher) *cobra.Command {
+	var (
+		inputsFlagVal   string
+		flagValsFlagVal string
+		dryRunFlagVal   bool
+	)
+	runPublishV2Cmd := &cobra.Command{
+		Use:   runPublishV2CmdName,
+		Short: "Runs the publish action for every product in the batch",
```
nmiyake:
```suggestion
		Short: "Runs the publish action for the provided products",
```

publisher/assetapi.go
```diff
@@ -126,6 +132,44 @@ func newRunPublishCmd(publisher distgo.Publisher) *cobra.Command {
...
+				return errors.Wrapf(err, "failed to unmarshal JSON %s", flagValsFlagVal)
+			}
+			return publisher.RunPublish(inputs, flagVals, dryRunFlagVal, cmd.OutOrStdout())
+		},
+	}
+	runPublishV2Cmd.Flags().StringVar(&inputsFlagVal, runPublishV2CmdInputsFlagName, "", "JSON representation of []distgo.PublishInput")
+	runPublishV2Cmd.Flags().StringVar(&flagValsFlagVal, runPublishCmdFlagValsFlagName, "", "JSON representation of map[distgo.PublisherFlag]any")
+	runPublishV2Cmd.Flags().BoolVar(&dryRunFlagVal, runPublishCmdDryRunFlagName, false, "true if the operation should be run as a dry run")
+	mustMarkFlagsRequired(runPublishV2Cmd,
+		runPublishV2CmdInputsFlagName,
```
nmiyake: nit

```suggestion
	mustMarkFlagsRequired(
	    runPublishV2Cmd,
		runPublishV2CmdInputsFlagName,
```

publisher/publisher_asset.go
nmiyake, line 74: Previously `args` was a separate variable basically just because of the `cfgYMLString == ""` check, but now that we're not doing that, you can actually just list all args inline here (or make `args :=` assignment a slice with all elements defined explicitly)

nmiyake (changes requested):
Thanks, this looks good for the most part! A few suggestions.

The type name change is just a suggestion/brainstorm -- if you would prefer to keep your naming or have a slightly different one that's fine.

Also, when the PR is next ready, can you confirm that you built it locally and tested using it with current assets? I know the test should cover this, but would like actual verification as well.

tabboud:
Thanks for the thorough review! I incorporated all the suggestions, so this should be ready for another review.

I tested this out locally ETE with the apollo publisher asset and ran a dry-run publish from a real product which worked as intended. I did the test in both directions, updated distgo and old asset and updated distgo with the updated asset and both worked correctly.

publisher/publisher_asset.go
tabboud: I measured this using a 3 product build with 4 os-arch combinations and one with a dependency on the other 2 and it came out to ~2.3KB. Then I tried some estimates using the 10/20/50/100 products and got the following sizes:
- 10 products: ~22.7KB
- 20 products: ~45.4KB
- 50 products: ~113.6KB
- 100 products: ~227.1KB

The relevant limit appears to be Linux's per-arg cap, `MAX_ARG_STRLEN` (128KB), since the whole batch goes into a single `--inputs` value. At this level I think the limit is ~55 products, so I think we're safe with the current approach. This number is also highly affected by the number of os-arch's and the dependencies since each product would embed all dependencies `ProductTaskOutputInfo`.

Here is a main.go file that I used to test the product sizes
```go
package main

import (
	"encoding/json"
	"fmt"

	"github.com/palantir/distgo/distgo"
)

func main() {
	// two independent products, auth-service, billing-service and one gateway-service that depends on both.
	// Each product has 4 os-archs.
	// A dependency embeds its dependency's full ProductOutputInfo (not just an ID).
	osArchDistInfo := &distgo.DistOutputInfos{
		DistOutputDir: "out/dist",
		DistIDs:       []distgo.DistID{"os-arch-bin"},
		DistInfos: map[distgo.DistID]distgo.DistOutputInfo{
			"os-arch-bin": {
				DistNameTemplateRendered: "gateway-service-1.234.0-12-gabcdef01234567890",
				DistArtifactNames: []string{
					"gateway-service-1.234.0-12-gabcdef01234567890-darwin-amd64.tgz",
					"gateway-service-1.234.0-12-gabcdef01234567890-darwin-arm64.tgz",
					"gateway-service-1.234.0-12-gabcdef01234567890-linux-amd64.tgz",
					"gateway-service-1.234.0-12-gabcdef01234567890-linux-arm64.tgz",
				},
				PackagingExtension: "tgz",
			},
		},
	}

	authService := distgo.ProductOutputInfo{
		ID:              "auth-service",
		Name:            "auth-service",
		DistOutputInfos: osArchDistInfo,
	}
	billingService := distgo.ProductOutputInfo{
		ID:              "billing-service",
		Name:            "billing-service",
		DistOutputInfos: osArchDistInfo,
	}
	gatewayService := distgo.ProductOutputInfo{
		ID:              "gateway-service",
		Name:            "gateway-service",
		DistOutputInfos: osArchDistInfo,
	}

	input := distgo.ProductPublishInfo{
		ProductTaskOutputInfo: distgo.ProductTaskOutputInfo{
			Project: distgo.ProjectInfo{
				ProjectDir: "/Volumes/git/go/src/github.p.b/some-org/realistic-service-name",
				Version:    "1.234.0-12-gabcdef01234567890",
			},
			Product: gatewayService,
			Deps: map[distgo.ProductID]distgo.ProductOutputInfo{
				"auth-service":    authService,
				"billing-service": billingService,
			},
		},
		PublisherConfigYML: []byte("group-id: com.palantir.some-org\nrepository: some-artifactory-repo\nurl: https://artifactory.p.b\n"),
	}

	single, err := json.Marshal(input)
	if err != nil {
		panic(err)
	}
	fmt.Printf("single ProductPublishInfo (4 os-archs, 2 deps): %d bytes (%.1f KB)\n", len(single), float64(len(single))/1024)

	for _, n := range []int{10, 20, 50, 100} {
		batch := make([]distgo.ProductPublishInfo, n)
		for i := range batch {
			batch[i] = input
		}
		b, err := json.Marshal(batch)
		if err != nil {
			panic(err)
		}
		fmt.Printf("%3d-product batch: %8d bytes (%.1f KB)\n", n, len(b), float64(len(b))/1024)
	}
}

```

nmiyake (approved):
Looks good, thanks!

PR #927, opened 2026-07-28
Fix: Batch publish across products sharing a GH release

tabboud:
## Before this PR
As of #923, `RunPublish` now allows publishers to coordinate publishing jobs across all products. Currently, when multiple products publish to the same GH release, the first products publish call will un-draft the release and when the immutable-releases feature is enabled, the second products publish call will fail since the release cannot be modified after finalizing. This was seen on golangci-lint-palantir (https://github.com/palantir/golangci-lint-palantir/actions/runs/29871701351/job/88773078087) since the initial product was finalized (making the release immutable) and the second product failed to finalize, resulting in a dangling draft with the second dist artifacts.

## After this PR

==COMMIT_MSG==
Batch publish across products sharing a GH release
==COMMIT_MSG==

After this change, the GH publisher will now batch products together based on the GH release they will create, creates or re-uses an existing draft release, uploads the assets for all products to that release, and then finalizes (un-drafts) the release once at the end.

## Possible downsides?

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/927)

### DIFF
--- a/publisher/github/publisher.go
+++ b/publisher/github/publisher.go
@@ -95,39 +95,6 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
 }
 
 func (p *githubPublisher) RunPublish(inputs []distgo.ProductPublishInfo, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
-	for _, input := range inputs {
-		if err := p.runPublish(input.ProductTaskOutputInfo, input.PublisherConfigYML, flagVals, dryRun, stdout); err != nil {
-			return errors.Wrapf(err, "failed to publish %s", input.ProductTaskOutputInfo.Product.ID)
-		}
-	}
-	return nil
-}
-
-func (p *githubPublisher) runPublish(productTaskOutputInfo distgo.ProductTaskOutputInfo, cfgYML []byte, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
-	var cfg config.GitHub
-	if err := yaml.Unmarshal(cfgYML, &cfg); err != nil {
-		return errors.Wrapf(err, "failed to unmarshal configuration")
-	}
-	if err := publisher.SetRequiredStringConfigValues(flagVals,
-		githubPublisherAPIURLFlag, &cfg.APIURL,
-		githubPublisherUserFlag, &cfg.User,
-		githubPublisherTokenFlag, &cfg.Token,
-		githubPublisherRepositoryFlag, &cfg.Repository,
-	); err != nil {
-		return err
-	}
-
-	if err := publisher.SetConfigValue(flagVals, githubPublisherOwnerFlag, &cfg.Owner); err != nil {
-		return err
-	}
-	if cfg.Owner == "" {
-		cfg.Owner = cfg.User
-	}
-
-	if err := publisher.SetConfigValue(flagVals, githubAddVPrefixFlag, &cfg.AddVPrefix); err != nil {
-		return err
-	}
-
 	filterRegexp, err := publisher.GetArtifactNamesFilterFlagValue(flagVals)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


publisher/github/publisher.go
```diff
@@ -261,7 +349,20 @@ func (p *githubPublisher) uploadFileAtPath(client *github.Client, release *githu
...
 	}
 
-	uploadURI, err := uploadURIForProduct(release.GetUploadURL(), path.Base(filePath))
+	assetName := path.Base(filePath)
+	if existingAsset := findReleaseAsset(release, assetName); existingAsset != nil {
+		stat, err := f.Stat()
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to stat artifact %s", filePath)
+		}
+		if int64(existingAsset.GetSize()) != stat.Size() {
```
nmiyake: I think in a previous version of this PR you mentioned that you were checking the size rather than the checksum because of the version of the GitHub API we were using or something like that -- would it be possible to update the API module version in this repo to check the checksum instead?

publisher/github/publisher.go
```diff
@@ -97,87 +97,183 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
...
+			return err
 		}
 	}
 	return nil
 }
 
-func (p *githubPublisher) runPublish(productTaskOutputInfo distgo.ProductTaskOutputInfo, cfgYML []byte, flagVals map[distgo.PublisherFlagName]any, dryRun bool, stdout io.Writer) error {
+// githubReleaseBatch groups the products that resolve to the same GitHub release so they can be uploaded together
+// and published once.
+type githubReleaseBatch struct {
```
nmiyake: I think moving the naming away from "Batch" makes it a bit clearer

```suggestion
// githubReleaseProducts stores the products that should be published for a particular GitHub release
type githubReleaseProducts struct {
```

publisher/github/publisher.go
```diff
@@ -97,87 +97,183 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
...
+	}
+	excludeRegexp, err := publisher.GetArtifactNamesExcludeFlagValue(flagVals)
+	if err != nil {
+		return err
+	}
+
+	// Group inputs by release key so that products sharing a release upload together and publish once, since a
+	// shared release must not be published until all of its products have uploaded.
+	var batches []githubReleaseBatch
+	batchIndexByKey := make(map[githubReleaseKey]int)
```
nmiyake:
```suggestion
	// Group inputs by release key so that products sharing a release upload together and publish once, since a
	// shared release must not be published until all of its products have uploaded.
	var releases []githubReleaseProducts
	releaseKeyToIndex := make(map[githubReleaseKey]int)
```

publisher/github/publisher.go
```diff
@@ -97,87 +97,183 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
...
+// githubReleaseBatch groups the products that resolve to the same GitHub release so they can be uploaded together
+// and published once.
+type githubReleaseBatch struct {
+	target   githubReleaseTarget
+	products []distgo.ProductTaskOutputInfo
+}
+
+// githubReleaseKey identifies a distinct GitHub release so that products resolving to the same one are grouped and
+// published together instead of once per product.
+type githubReleaseKey struct {
```
nmiyake, line 177: This is probably safe, but technically `Token` is also a config-derived value that could theoretically be different between products. This would probably be rare in practice since it would require a product to define its token as part of its config, but our current structure does allow that.

I think the behavior if the token is different is not well-defined in the current model (since even if we could use a client with a different token for each product that specifies it as such, in that case it's not clear what token to use for the over-all publish operation), so probably okay to just check for this and report an error if there are products that have config that differ only in token between githubReleaseKey structs

publisher/github/publisher.go
```diff
@@ -97,87 +97,183 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
...
+		cfg, key, err := resolveGitHubReleaseConfig(input.PublisherConfigYML, flagVals, productTaskOutputInfo.Project.Version)
+		if err != nil {
+			return errors.Wrapf(err, "failed to resolve GitHub config for product %s", productTaskOutputInfo.Product.ID)
+		}
+
+		batchIndex, ok := batchIndexByKey[key]
+		if !ok {
+			// Only the first product seen for a given release key needs a client. Every other product in the
+			// batch reuses this target instead of re-resolving.
+			target, err := newGitHubReleaseTarget(cfg, key)
```
nmiyake:
```suggestion
			// if release target does not exist for the key, create it
			target, err := newGitHubReleaseTarget(cfg, key)
```

publisher/github/publisher.go
```diff
@@ -97,87 +97,183 @@ func (p *githubPublisher) Flags() ([]distgo.PublisherFlag, error) {
...
+			}
+			batchIndex = len(batches)
+			batchIndexByKey[key] = batchIndex
+			batches = append(batches, githubReleaseBatch{target: target})
+		}
+		batches[batchIndex].products = append(batches[batchIndex].products, productTaskOutputInfo)
+	}
+
+	// With grouping done, run the core prepare/upload/publish workflow once per batch. For each batch, create or
+	// reuse the release, upload every product's assets in the batch, then publish it for the whole batch at once.
```
nmiyake:
```suggestion
	// With grouping done, run the core prepare/upload/publish workflow once per release. For each release, create or
	// reuse the release, upload every product's assets to the release, then publish it.
```

nmiyake (changes requested):
Thanks! Core looks good, but some feedback.

publisher/github/publisher.go
tabboud: Yes good recall. I opened https://github.com/palantir/distgo/pull/928 which updates the go-github version to one that supports the release asset digest API. That can then be used instead of this size based check. I'll wait for that to merge before updating this PR.

publisher/github/publisher.go
tabboud: I considered adding the `Token` initially but assumed it would always be the same across products publishing to the same release. Instead of adding the `Token` to the key I opted for explicitly checking this case and returning an error if a single release contains differing tokens.

tabboud:
@nmiyake Thanks for the feedback! Everything should be addressed now and the latest changes include checksum verification with a fallback to a size check in case the checksum/digest is not supported by the version of GH being queried.

nmiyake (approved):
Looks good, thank you!

PR #937, opened 2026-08-05
Add Docker output directory

bmoylan:
Rebases and expands upon https://github.com/palantir/distgo/pull/428 and allows Docker-only builds without a corresponding dist.

- Adds a configurable `docker.output-dir`, defaulting to `out/docker`, for OCI layouts and other Docker build outputs.
- Resolves that directory in the plugin and passes it to the DockerBuilder, which writes to `ProductDockerOutputDirCandidates(...)[0]` — the location every distgo version that could be driving the build agrees on — instead of deriving the path from its own version.
- Preserves compatibility with existing vendored DockerBuilder assets by recognizing their legacy `out/dist/.../oci-<docker-id>` layouts during wrapper creation, dependency resolution, and push; a builder driven by a distgo that predates the output directory writes to the legacy location rather than into the source tree.
- Removes a stale legacy layout before a build, so `docker push` cannot publish an image from an earlier build at the same version.
- Removes the Docker output tree during `dist clean`.
- Rejects Docker output directories that are absolute, the project directory itself, or outside it, since `clean` removes `<output-dir>/<product-id>` wholesale.

A DockerBuilder can be a separate asset binary that vendors distgo independently of the plugin, so the two are routinely at different versions. No upgrade ordering is required: the plugin searches both the new and the legacy location, and an asset writes to whichever one the plugin driving it looks in, since that plugin supplies the candidate list it is handed.

Closes #428

- - -
This change is [Reviewable](https://reviewable.io/reviews/palantir/distgo/937)

### DIFF
--- a/distgo/docker/docker_push.go
+++ b/distgo/docker/docker_push.go
@@ -95,16 +95,17 @@ func runSingleDockerPush(
 	stdout io.Writer) (rErr error) {
 
 	// if an OCI artifact exists, push that. Otherwise, default to pushing the artifact in the docker daemon
-	if _, err := layout.FromPath(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID)); err == nil {
+	if _, err := layout.FromPath(distgo.ProductDockerOutputDir(productTaskOutputInfo.Project, productTaskOutputInfo.Product, dockerID)); err == nil {
 		return runOCIPush(productID, dockerID, productTaskOutputInfo, dryRun, insecure, stdout)
 	}
 	return runDockerDaemonPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
 }
 
 func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, insecure bool, stdout io.Writer) error {
-	index, err := layout.ImageIndexFromPath(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID))
+	outputDir := distgo.ProductDockerOutputDir(productTaskOutputInfo.Project, productTaskOutputInfo.Product, dockerID)
+	index, err := layout.ImageIndexFromPath(outputDir)
 	if err != nil {
-		return errors.Wrapf(err, "failed to construct image index from OCI layout at path %s", productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID))
+		return errors.Wrapf(err, "failed to construct image index from OCI layout at path %s", outputDir)
 	}
 
 	for _, tag := range productTaskOutputInfo.Product.DockerOutputInfos.DockerBuilderOutputInfos[dockerID].RenderedTags {
@@ -185,7 +186,7 @@ func handleImageIndex(index v1.ImageIndex, idxManifest *v1.IndexManifest, ref na
 }
 
 func handleImageManifest(ref name.Reference, productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, stdout io.Writer) error {
-	path := filepath.Join(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID), "image.tar")
+	path := filepath.Join(distgo.ProductDockerOutputDir(productTaskOutputInfo.Project, productTaskOutputInfo.Product, dockerID), "image.tar")
 	image, err := tarball.ImageFromPath(path, nil)
 	if err != nil {
 		return errors.Wrapf(err, "failed to read image from path %s", path)
--- a/distgo/docker/docker_push.go
+++ b/distgo/docker/docker_push.go
@@ -95,14 +95,22 @@ func runSingleDockerPush(
 	stdout io.Writer) (rErr error) {
 
 	// if an OCI artifact exists, push that. Otherwise, default to pushing the artifact in the docker daemon
-	if _, err := layout.FromPath(distgo.ProductDockerOutputDir(productTaskOutputInfo.Project, productTaskOutputInfo.Product, dockerID)); err == nil {
-		return runOCIPush(productID, dockerID, productTaskOutputInfo, dryRun, insecure, stdout)
+	if outputDir := dockerOCIOutputDir(productTaskOutputInfo, dockerID); outputDir != "" {
+		return runOCIPush(productID, dockerID, productTaskOutputInfo, outputDir, dryRun, insecure, stdout)
 	}
 	return runDockerDaemonPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
 }
 
-func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, insecure bool, stdout io.Writer) error {
-	outputDir := distgo.ProductDockerOutputDir(productTaskOutputInfo.Project, productTaskOutputInfo.Product, dockerID)
+func dockerOCIOutputDir(productTaskOutputInfo distgo.ProductTaskOutputInfo, dockerID distgo.DockerID) string {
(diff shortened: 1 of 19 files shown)


distgo/product.go
```diff
@@ -134,27 +134,78 @@ func ProductBuildArtifactPaths(projectInfo ProjectInfo, productOutputInfo Produc
...
+	return ProductDistOutputDir(projectInfo, productOutputInfo, DistID("oci-"+dockerID))
+}
+
+// ProductDockerOCIDistOutputDir returns the legacy Docker OCI dist output directory for the given DockerID.
 //
-// Note that this scheme uses the namespace for dist outputs, so if a product has a DockerID "X" and a dist with
-// DistID "oci-X", then the output directories will be the same and the behavior will be undefined -- this is a
-// known issue/risk that we are accepting as part of the design.
+// Deprecated: use ProductDockerOutputDirCandidates. This method retains its old result so DockerBuilder assets compiled
+// against older distgo versions continue to interoperate with the host during migration to the Docker output directory.
```
nmiyake:
```suggestion
// ProductDockerOCIDistOutputDir returns the legacy Docker OCI dist output directory for the given DockerID, which is
// "{{ProjectDir}}/{{DistOutputDir}}/{{ProductID}}/{{Version}}/oci-{{DockerID}}".
//
// Deprecated: use ProductDockerOutputDirCandidates. This method retains its old result so DockerBuilder assets compiled
// against older distgo versions continue to interoperate with the host during migration to the Docker output directory.
```

distgo/paramdocker.go
```diff
@@ -41,11 +41,20 @@ type DockerParam struct {
 	// Repository is the Docker repository. This value is made available to TagTemplates as {{Repository}}.
 	Repository string
 
+	// OutputDir specifies the default output directory for any on-disk artifacts created by the "docker build" task --
+	// for example, OCI image layouts, metadata, etc. Different from the ContextDir in that the ContextDir is used as
+	// the location in which inputs are placed and where images are built, whereas the OutputDir dictates where any
+	// output may be written. Not all Docker build tasks have output, so the output directory may not be created or may
+	// be empty even on successful executions. The artifacts for a Docker configuration are written to
+	// "{{OutputDir}}/{{ID}}/{{Version}}/{{DockerID}}".
```
nmiyake:
```suggestion
	// "{{OutputDir}}/{{ProductID}}/{{Version}}/{{DockerID}}".
```

distgo/docker/docker_push.go
```diff
@@ -95,16 +95,25 @@ func runSingleDockerPush(
...
-		return runOCIPush(productID, dockerID, productTaskOutputInfo, dryRun, insecure, stdout)
+	if outputDir := dockerOCIOutputDir(productTaskOutputInfo, dockerID); outputDir != "" {
+		return runOCIPush(productID, dockerID, productTaskOutputInfo, outputDir, dryRun, insecure, stdout)
 	}
 	return runDockerDaemonPush(productID, dockerID, productTaskOutputInfo, dryRun, stdout)
 }
 
-func runOCIPush(productID distgo.ProductID, dockerID distgo.DockerID, productTaskOutputInfo distgo.ProductTaskOutputInfo, dryRun bool, insecure bool, stdout io.Writer) error {
-	index, err := layout.ImageIndexFromPath(productTaskOutputInfo.ProductDockerOCIDistOutputDir(dockerID))
+func dockerOCIOutputDir(productTaskOutputInfo distgo.ProductTaskOutputInfo, dockerID distgo.DockerID) string {
```
nmiyake, line 107: It would be helpful to have documentation on this function, particularly around when the return value may be empty and what happens when the return value is empty.

distgo/paramdocker.go
nmiyake: I know this is set to a default value by config if the config is empty, but Param values can be constructed in-memory as well. It looks like this value being empty has certain implications on behavior, so I think it would be helpful to document that here.

distgo/product.go
```diff
@@ -134,27 +134,78 @@ func ProductBuildArtifactPaths(projectInfo ProjectInfo, productOutputInfo Produc
...
+// ProductDockerOutputDir returns the output directory for the docker outputs for the docker builder with the given
+// DockerID, which is "{{ProjectDir}}/{{DockerOutputDir}}/{{ProductID}}/{{Version}}/{{DockerID}}".
+func ProductDockerOutputDir(projectInfo ProjectInfo, productOutputInfo ProductOutputInfo, dockerID DockerID) string {
+	if productOutputInfo.DockerOutputInfos == nil {
+		return ""
+	}
+	relDir := ProductDockerOutputRelDir(productOutputInfo.DockerOutputInfos.DockerOutputDir, productOutputInfo.ID, projectInfo.Version, dockerID)
+	if relDir == "" {
+		return ""
+	}
```
nmiyake, line 156: When can/does this path happen?

distgo/product.go
```diff
@@ -134,27 +134,78 @@ func ProductBuildArtifactPaths(projectInfo ProjectInfo, productOutputInfo Produc
...
+		return nil
+	}
+	var outputDirs []string
+	addDir := func(outputDir string) {
+		if outputDir != "" && !slices.Contains(outputDirs, outputDir) {
+			outputDirs = append(outputDirs, outputDir)
+		}
+	}
+	if relDir := productOutputInfo.DockerOutputInfos.DockerBuilderOutputInfos[dockerID].OutputDir; relDir != "" {
+		addDir(path.Join(projectInfo.ProjectDir, relDir))
```
nmiyake, line 189: Is there a scenario where a properly configured product/project can return an empty `relDir` here, and thus not add this entry?

dockerbuilder/defaultdockerbuilder/buildcontext.go
```diff
@@ -49,13 +49,13 @@ func dependencyImageBuildContextArgs(productTaskOutputInfo distgo.ProductTaskOut
...
-			if ociDir == "" {
+			outputDirs := distgo.ProductDockerOutputDirCandidates(productTaskOutputInfo.Project, depOutputInfo, dockerID)
+			if len(outputDirs) == 0 {
 				continue
 			}
-			layoutDir := filepath.Join(ociDir, distgo.DockerBuildContextLayoutSubdir)
+			layoutDir := filepath.Join(outputDirs[0], distgo.DockerBuildContextLayoutSubdir)
 			if !dryRun {
-				if _, err := os.Stat(filepath.Join(layoutDir, "index.json")); err != nil {
+				if layoutDir = existingLayoutDir(outputDirs); layoutDir == "" {
```
nmiyake: Hmm, not sure if this can be helped, but the updated logic is slightly different from the previous in that `existingLayoutDir` could change the value of `layoutDir` -- in the previous implementation, `layoutDir` would always have the same value (but in the "real" run, it could be skipped), whereas now if this call returns non-empty, it may modify the value of `layoutDir` and thus make the actual run different from what was reported in `dry-run`

nmiyake (changes requested):
Thanks! In sync with the core of the change, but some questions/suggestions on documentation.

distgo/paramdocker.go
bmoylan: 👍 added a mention that empty means we have nowhere to write

distgo/product.go
bmoylan: Documented on the function, but an old asset or a DockerParam constructed in memory might be empty

distgo/product.go
bmoylan: ToProductOutputInfo always resolves it for assets that update their dependency but it's empty when the output info came from an older plugin that has no such field to send

dockerbuilder/defaultdockerbuilder/buildcontext.go
bmoylan: Good catch. Fixed by making both modes select the same way

bmoylan:
@nmiyake Thanks for the review! I think I have addressed everything

nmiyake (approved):
Looks good -- thank you!

bmoylan:
👍

